PR #51, opened 2026-03-17
chore: remove notion-sync and clean up Notion references

knoxio:
## Summary
- Remove `apps/notion-sync/` service entirely (1800+ lines of code)
- Remove Notion-related scripts (`cleanup-all-duplicates.ts`, `cleanup-duplicate-entities.ts`, `sync-entities-only.ts`)
- Clean up Notion references in CLAUDE.md, README.md, docker-compose, Ansible templates
- Remove notion-sync CI workflow
- Update db-types to remove `notion_id` fields
- Clean up `.env.example` and `package.json` workspace references

## Test plan
- [ ] Verify `mise typecheck` passes
- [ ] Verify `mise test` passes
- [ ] Verify docker-compose config is valid
- [ ] Verify no remaining Notion runtime dependencies

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -181,7 +181,7 @@ ssh -p 2222 pops@pops.local
 cd /opt/pops
 docker compose logs -f finance-api
 docker compose logs -f pops-pwa
-docker compose logs -f moltbot
+docker compose logs -f metabase
 
 # Check service health
 docker compose ps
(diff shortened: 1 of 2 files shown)


knoxio:
## Review response

All issues addressed:

**Must fix:**
1. **Ansible init command** — Added `register: db_init_result` and a follow-up `Report database initialization failure` task that fails with stderr output when rc != 0. The `failed_when: false` on the check task is correct: rc=0 (db exists), rc=1 (doesn't exist), rc>1 (Docker error, caught by the "Fail if Docker is not available" task). `async: 120 / poll: 10` was already present.
2. **No timeout** — Already had `async: 120 / poll: 10` (see above).

**Minor:**
- **`imports/service.ts` comment** — Fixed to: "Also insert into SQLite for local queries (Notion is still the write target above)"
- **`docs/DEPLOYMENT.md` log example** — Changed from `moltbot` to `metabase` (more useful for debugging than moltbot).

Pushed: 1b9def6

PR #52, opened 2026-03-17
feat(db): replace notion_id primary keys with auto-generated UUIDs

knoxio:
## Summary
- Replaces Notion-specific `notion_id` primary keys with auto-generated UUIDs across all database tables
- Updates all queries, types, and references to use the new `id` column
- Enables the database to function independently of Notion as the source of truth

## Test plan
- [ ] Verify all existing tests pass with new UUID primary keys
- [ ] Confirm database migrations apply cleanly on fresh init
- [ ] Check that seed data generates valid UUIDs

### DIFF
--- a/apps/finance-api/src/db/schema.ts
+++ b/apps/finance-api/src/db/schema.ts
@@ -26,6 +26,10 @@ const INCLUDED_MIGRATIONS = [
  * Initialize a fresh SQLite database with the full POPS schema.
  * Safe to call on an empty file or an already-initialized database
  * (all statements use CREATE TABLE IF NOT EXISTS).
+ *
+ * Note: SQLite DEFAULTs use `lower(hex(randomblob(16)))` (32-char hex) as a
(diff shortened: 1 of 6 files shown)


knoxio:
## Review: APPROVE with nits

**Approach:** Correct use of CREATE TABLE + copy + DROP + RENAME for SQLite PK changes. Existing notion_id values preserved as both `id` and `notion_id` for backward compatibility. Clean removal of `@ts-expect-error` comments by extracting proper `NotionCreateProperties`/`NotionUpdateProperties` types.

**Nits:**
- `lower(hex(randomblob(16)))` produces 32-char hex strings, not RFC 4122 UUIDs. `crypto.randomUUID()` in service code produces proper UUIDs — two different ID formats. Consider using a consistent format.
- Tests removed `expect(budget).not.toHaveProperty("notion_id")` assertions — those were still useful since `notion_id` remains a column and shouldn't leak into API responses.

PR #123, opened 2026-03-20
feat: TheTVDB metadata refresh [PRD-009/US-5] (tb-067)

knoxio:
## Summary
- Add `refreshTvShow()` to `thetvdb/service.ts` — fetches fresh show detail from TheTVDB, updates local metadata (preserving `poster_override_path`), upserts seasons/episodes without deleting existing records
- Add `refreshTvShow` tRPC mutation to TV shows router with input `{ id, redownloadImages?, refreshEpisodes? }`
- Image re-download stubbed pending image cache service (tb-065)

## Test plan
- [x] 10 unit tests covering:
  - Updates show metadata from TheTVDB
  - Preserves `poster_override_path` on refresh
  - Inserts new seasons and episodes
  - Updates existing episodes without deleting any
  - Handles show with no seasons
  - Handles specials (season 0)
  - Throws NOT_FOUND for invalid show id
  - Skips episode refresh when `refreshEpisodes: false`
  - Updates genres and networks from fresh data
  - Continues refreshing other seasons when one season's episode fetch fails
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes (653 tests)
- [x] Lint clean on changed files

### DIFF
--- a/apps/pops-api/src/modules/media/index.ts
+++ b/apps/pops-api/src/modules/media/index.ts
@@ -1,5 +1,5 @@
 /**
- * Media domain — movies, tv shows, comparisons, watchlist, watch history.
+ * Media domain — movies, tv shows, comparisons, watchlist, watch history, library.
  */
 import { router } from "../../trpc.js";
 import { moviesRouter } from "./movies/router.js";
--- a/apps/pops-api/src/modules/media/library/router.ts
+++ b/apps/pops-api/src/modules/media/library/router.ts
@@ -14,6 +14,7 @@ import { RefreshMovieSchema } from "./types.js";
 import * as libraryService from "./service.js";
 import { getTvdbClient } from "../thetvdb/index.js";
 import { TvdbApiError } from "../thetvdb/types.js";
+import { refreshTvShow } from "../thetvdb/service.js";
 import * as tvShowService from "./tv-show-service.js";
 
 /** Shared rate limiter: TMDB allows 40 req / 10 s → 4 req/s. */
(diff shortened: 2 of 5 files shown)


knoxio (commented):
## QA Review: TheTVDB metadata refresh [PRD-009/US-5]

### Build Status
- **Typecheck**: PASS (all 7 tasks)
- **Tests**: PASS (653 API + 23 shell = 676 total, 0 failures)

---

### Blocking Issues

**(none)**

---

### Should-Fix Issues

**1. New TheTvdbAuth + TheTvdbClient created on every request (router.ts:210-211)**

Every call to `refreshTvShow` creates a brand new `TheTvdbAuth` and `TheTvdbClient` instance. Since `TheTvdbAuth` caches the JWT token *on the instance*, this means every refresh request triggers a fresh POST /login to TheTVDB before doing the actual work. The token lasts ~1 month, so this is wasteful.

The TMDB client takes a rate limiter in its constructor, suggesting the project expects clients to be long-lived singletons. Consider creating a module-level singleton (lazy-initialized), e.g.:

```ts
let sharedClient: TheTvdbClient | null = null;
function getTheTvdbClient(): TheTvdbClient {
  if (!sharedClient) {
    const apiKey = process.env.THETVDB_API_KEY;
    if (!apiKey) throw new TRPCError({ code: "INTERNAL_SERVER_ERROR", message: "THETVDB_API_KEY not configured" });
    sharedClient = new TheTvdbClient(new TheTvdbAuth(apiKey));
  }
  return sharedClient;
}
```

**2. `numberOfSeasons` counts specials (service.ts:58)**

`numberOfSeasons: detail.seasons.length` includes season 0 (Specials) when present. For Breaking Bad with specials, this would report 6 seasons instead of 5. Filter out season 0:

```ts
numberOfSeasons: detail.seasons.filter(s => s.seasonNumber > 0).length,
```

**3. Episode `stillPath` / `imageUrl` dropped during upsert (service.ts:145-148, 157-166)**

The `TvdbEpisode` type has an `imageUrl` field, and the `episodes` table has a `stillPath` column, but the upsert code never writes it. Episode images from TheTVDB are silently lost. Both the insert and update paths should include:

```ts
// In update .set():
stillPath: ep.imageUrl,
// In insert .values():
stillPath: ep.imageUrl,
```

**4. No transaction wrapping for multi-table upserts (service.ts:70-181)**

The `refreshEpisodes` block performs multiple inserts/updates across `seasons` and `episodes` tables in a loop with no transaction wrapping. If the process crashes mid-way (e.g., network error fetching season 3's episodes after seasons 1-2 are committed), the database is left in a partially-updated state. Consider wrapping the entire block in a Drizzle transaction (or a raw SQLite transaction via `db.transaction()`).

**5. Extra SELECT after INSERT to get season ID (service.ts:109-117)**

After inserting a new season, the code does a separate SELECT to retrieve its ID. Drizzle supports `.returning()` on SQLite inserts, which avoids the round-trip:

```ts
const [newSeason] = db.insert(seasons).values({...}).returning({ id: seasons.id }).all();
seasonId = newSeason.id;
```

---

### Minor Issues

**6. Variable name reuse: `g` in networks map (service.ts:60)**

`networks: detail.networks.map((g) => g.name)` uses `g` as the parameter name, but these are networks, not genres. Use `n` for clarity:

```ts
networks: detail.networks.map((n) => n.name),
```

**7. Test mock uses `as unknown as TheTvdbClient` (service.test.ts:42, 415)**

Two test helpers cast partial mocks via `as unknown as TheTvdbClient`. This is fine for now but could mask future type mismatches if the client interface changes. Consider extracting a minimal interface (e.g. `TheTvdbClientLike`) that the service depends on, making the mock naturally type-safe.

---

### What Looks Good
- Clean separation: service is pure logic with injected client, router handles wiring
- Tests are thorough (10 tests covering happy path, edge cases, error resilience)
- Upsert-only strategy (no deletes) is safe and correct
- `poster_override_path` preservation is tested
- Error handling for per-season episode fetch failures is well-designed (continue to next season)
- No `any` types in production code
- All exports properly wired through `thetvdb/index.ts`

---

### Recommendation: **Request Changes**

Issues #1-4 should be addressed before merge. The core logic is solid but the missing transaction wrapping (#4) and per-request auth (#1) are the most important fixes. The `stillPath` omission (#3) means data is silently lost, and the specials count (#2) is a correctness bug.

PR #147, opened 2026-03-20
fix: resolve lint/type errors blocking CI on main [tb-091]

knoxio:
## Summary
- Add missing `@pops/api` and `@trpc/react-query` dependencies to `app-media` package — resolves cascading type errors (`Cannot find module`, implicit `any` parameters)
- Move `formatRuntime` to shared `utils.ts` and import in both `MovieDetailPage` and `EpisodeList` — fixes `has no exported member` error
- Add `eslint` + flat config to `pops-shell` — lint script existed but `eslint` was never installed
- Update lint scripts to flat config syntax (drop deprecated `--ext` flag)

## Test plan
- [x] `pnpm typecheck` passes for both pops-api and pops-shell
- [x] `pnpm lint` passes for pops-shell (was previously failing with `eslint: not found`)
- [x] All 792 backend tests pass
- [x] Shell builds successfully


knoxio:
CI needs to re-run — lint fixes landed in 818761c and 7afc4b1 but checks are stale. Lint passes locally.

Issue #733, opened 2026-03-26
Complete PRD-043 US-04: Connections and photos API


knoxio:
## What exists

tRPC procedures exist for connections (`connect`, `listForItem`, `traceChain` with maxDepth and cycle handling) and photos (`delete`, `reorder`, `listForItem`). The connect procedure normalises A<B ordering and handles self-connection/duplicate rejection. traceChain uses recursive CTE with visited-items set.

## What's missing

- Implement `inventory.connections.disconnect` procedure — accepts two item IDs, normalises A<B ordering, deletes the connection row
- Implement `inventory.photos.upload` procedure — accepts itemId and multipart file, compresses image (1920px max dimension, HEIC to JPEG conversion, strip EXIF metadata), stores file at `{INVENTORY_IMAGES_DIR}/items/{itemId}/photo_NNN.jpg`, creates database record
- Ensure photo upload creates the item's directory if it does not exist
- Implement sequential photo filenames within an item's directory (photo_001.jpg, photo_002.jpg, etc.)
- Write tests covering connect/disconnect with A<B normalisation, duplicate/self-connection rejection, listForItem queries both sides, traceChain recursive traversal and max depth limit, photo upload with compression, photo delete (record + file), photo reorder

## Acceptance criteria

- [ ] `inventory.connections.disconnect` — accepts two item IDs, normalises ordering, deletes the connection row
- [ ] `inventory.photos.upload` — accepts itemId and file (multipart), compresses image (1920px max dimension, HEIC→JPEG conversion, strip EXIF metadata), stores file at `{INVENTORY_IMAGES_DIR}/items/{itemId}/photo_NNN.jpg`, creates database record
- [ ] Photo upload creates the item's directory if it does not exist
- [ ] Photo filenames are sequential within an item's directory (photo_001.jpg, photo_002.jpg, etc.)
- [ ] Tests cover: connect/disconnect with A<B normalisation, duplicate connection rejection, self-connection rejection, listForItem queries both sides, traceChain recursive traversal, traceChain max depth limit, photo upload with compression, photo delete (record + file), photo reorder

## Files

- US: `docs-v2/themes/04-inventory/prds/043-inventory-data-model-api/us-04-connections-photos-api.md`

PR #832, opened 2026-03-26
feat: connections disconnect by items + photo upload with compression [GH-733]

knoxio:
## Summary
- Add `inventory.connections.disconnectByItems` procedure — accepts two item IDs, normalises A<B ordering, deletes the connection row
- Add `inventory.photos.upload` procedure — accepts base64-encoded image, compresses with sharp (1920px max, HEIC→JPEG, strip EXIF), stores with sequential filenames (`photo_001.jpg`, etc.), creates item directory if missing
- Add `INVENTORY_IMAGES_DIR` env var (default: `./data/inventory/images`)
- Install sharp dependency for image processing
- Update photo `removePhoto` to also delete the file from disk
- Add 7 trace tests covering recursive traversal, maxDepth limit, circular reference handling, and node metadata
- Add photo delete file removal test (verifies both DB record and disk file are removed)

## Test plan
- [x] 3 tests for `disconnectByItems` (normal, reversed order, not found)
- [x] 4 tests for `photos.upload` (compress, sequential filenames, not found, auth)
- [x] 7 tests for `connections.trace` (tree structure, recursive chain, maxDepth, circular refs, empty, not found, metadata)
- [x] 1 test for photo delete file removal (upload → remove → verify file deleted)
- [x] All 173 inventory tests pass (7 test files)
- [x] Typecheck clean (only pre-existing msw module error)

### DIFF
--- a/apps/pops-api/src/modules/inventory/connections/connections.test.ts
+++ b/apps/pops-api/src/modules/inventory/connections/connections.test.ts
@@ -406,6 +406,152 @@ describe("inventory.connections.graph", () => {
   });
 });
 
+describe("inventory.connections.trace", () => {
+  it("returns a tree rooted at the starting item", async () => {
+    const idA = seedInventoryItem(db, { item_name: "Hub" });
+    const idB = seedInventoryItem(db, { item_name: "Device 1" });
+    const idC = seedInventoryItem(db, { item_name: "Device 2" });
+
+    const pairAB = [idA, idB].sort() as [string, string];
+    const pairAC = [idA, idC].sort() as [string, string];
+    seedItemConnection(db, pairAB[0], pairAB[1]);
+    seedItemConnection(db, pairAC[0], pairAC[1]);
+
+    const result = await caller.inventory.connections.trace({ itemId: idA });
+
+    expect(result.data.id).toBe(idA);
+    expect(result.data.itemName).toBe("Hub");
+    expect(result.data.children).toHaveLength(2);
+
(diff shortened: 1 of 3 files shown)


knoxio (commented):
## Review — PR #832: Connections disconnect by items + photo upload with compression

Good feature PR. The `disconnectByItems` and `uploadPhoto` implementations are solid, with good test coverage. However, **CI is failing** — lint error needs fixing before merge.

### CI Failure: Lint
`photos/service.ts:178` — `match[1]!` uses a non-null assertion forbidden by `@typescript-eslint/no-non-null-assertion`.

Fix:
```ts
// Before
const num = parseInt(match[1]!, 10);
// After
const num = parseInt(match[1] ?? "0", 10);
```

### Code looks good otherwise
- `disconnectByItems` correctly normalizes A<B ordering — consistent with `connectItems`
- Photo upload: sharp compression pipeline (rotate→resize→removeAlpha→jpeg) is correct for HEIC→JPEG
- Sequential filename logic via `readdirSync` + regex scan — simple and correct
- `mkdirSync({ recursive: true })` handles missing directories
- Tests use `mkdtempSync` + `vi.stubEnv` for isolated temp directories — good practice
- Auth tests included for upload endpoint

### Minor observation (not blocking)
- `compressImage` does `sharp(buffer).toBuffer()` then `sharp(compressed).toFile(path)` — could be a single pipeline with `.toFile()` directly. But the two-step approach is fine for correctness and the buffer is needed for the intermediate step.

**Requesting changes** — fix the lint error (`match[1]!` → `match[1] ?? "0"`) and push to get CI green.

PR #834, opened 2026-03-26
feat: per-section error states and result cap for media search

### LINKED ISSUE #719: Complete PRD-032 US-02: Search results display
## What exists

Results display in two sections: "Movies" and "TV Shows". Each result card shows poster thumbnail, title, release year, and truncated overview. Poster thumbnails use fallback placeholders. Each section has independent loading state and empty state messages. "In Library" detection works by checking against existing tmdbIds/tvdbIds.

## What's missing

- Error state is a single combined state covering both sections instead of per-section. Each section should have its own independent error state with a "Retry" button that re-fires only that section's API call (e.g., retry TMDB search without re-firing TheTVDB search).
- No per-section result cap enforced. Results should be capped at a reasonable limit (e.g., 20 per section) matching API response size, with the result list scrollable.
- Tests are missing for: both sections rendering independently, poster fallback, "In Library" badge for existing items, per-section loading/empty/error states, overview truncation.

## Acceptance criteria

- [ ] Each section has its own error state with a "Retry" button that re-fires only that section's API call
- [ ] Result list is scrollable; results are capped at a reasonable limit (e.g., 20 per section) matching API response size
- [ ] Tests cover: both sections render independently, poster fallback, "In Library" badge appears for existing items, per-section loading/empty/error states, overview truncation

## Files

- US: `docs-v2/themes/03-media/prds/032-search-page/us-02-search-results.md`

knoxio:
## Summary
- Replace combined error state with per-section inline error + retry UI (Movies and TV independently)
- Each section's Retry button only re-fires that section's API call
- Cap results at 20 per section via `.slice(0, 20)`
- Per-section loading skeletons (3 per section instead of 6 combined)
- Remove combined error toast `useEffect` — inline errors are more visible

Closes #719

## Test plan
- [ ] Search for a query — both sections render independently
- [ ] If movie search fails (e.g., network error), only movie section shows error with Retry button
- [ ] Clicking Retry on movie error only refetches movies, not TV
- [ ] Results are capped at 20 per section
- [ ] Each section shows its own loading skeletons independently

### DIFF
--- a/docs-v2/themes/03-media/prds/032-search-page/us-02-search-results.md
+++ b/docs-v2/themes/03-media/prds/032-search-page/us-02-search-results.md
@@ -1,7 +1,7 @@
 # US-02: Search results display
 
 > PRD: [032 — Search Page](README.md)
-> Status: Partial
+> Status: Partial — tests outstanding
 
 ## Description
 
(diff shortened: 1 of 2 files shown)


knoxio (commented):
## Review — PR #834: Per-section error states and result cap

Good refactor — replacing combined error state with independent per-section error/loading/results is the right pattern for a multi-source search page.

- Per-section inline error with retry button — clean UX, shows the actual error message
- Result cap at 20 via `.slice(0, MAX_RESULTS_PER_SECTION)` — simple and correct
- Skeleton count reduced from 6 to 3 per section — more realistic preview
- Removed `useEffect` error toasts in favor of inline errors — better discoverability
- `noResults` logic updated to use `moviesSettled && tvSettled` — correctly handles mixed error/empty states
- Section headers show count only when results exist — good detail

Minor observation (not blocking): the skeleton markup is duplicated between movie and TV sections. Could extract a `SearchSkeleton` component, but it's fine for now.

**Approved** — looks good to merge.

PR #1000, opened 2026-03-28
fix(media): prevent Arr settings wipe when saving one service

### LINKED ISSUE #979: Arr settings: saving one service wipes unsaved changes in the other
## Bug

On the Arr Settings page, if you enter details for both Radarr and Sonarr and then save only one, the other service's unsaved form values get wiped.

## Root Cause

In `packages/app-media/src/pages/ArrSettingsPage.tsx`:

1. User fills in both Radarr and Sonarr fields (local state only)
2. User clicks Save on Radarr — mutation sends only `radarrUrl` + `radarrApiKey` (lines 227-230)
3. `onSuccess` callback calls `settingsQuery.refetch()` (line 161)
4. The `useEffect` on line 148-156 fires when refetch completes, resetting ALL local state (including Sonarr fields) from the server response
5. Since Sonarr was never saved, the server returns empty/previous values, wiping the user's input

## Fix Options

1. **Don't refetch on save** — just update local state optimistically, or only refetch the saved service's fields
2. **Track dirty state per service** — skip overwriting fields that have unsaved local changes during the useEffect
3. **Send all fields on save** — include both services' current local state in every save mutation call

## Files

- `packages/app-media/src/pages/ArrSettingsPage.tsx:148-164,226-231,246-251`

knoxio:
## Summary
- Removed `settingsQuery.refetch()` from save `onSuccess` callback
- The refetch was triggering the `useEffect` to reset ALL field state, wiping unsaved changes in the other service (e.g., saving Radarr would wipe unsaved Sonarr fields)
- Local state already holds the correct values after save; initial load `useEffect` handles first-time population

Fixes #979

## Test plan
- [ ] Enter Radarr URL + key, then enter Sonarr URL + key (don't save yet)
- [ ] Save Radarr — Sonarr fields should still have their unsaved values
- [ ] Save Sonarr — both services should be saved correctly
- [ ] Refresh page — both saved values should load correctly

### DIFF
--- a/packages/app-media/src/pages/ArrSettingsPage.tsx
+++ b/packages/app-media/src/pages/ArrSettingsPage.tsx
@@ -158,7 +158,6 @@ export function ArrSettingsPage() {
   const saveSettings = trpc.media.arr.saveSettings.useMutation({
     onSuccess: () => {
       toast.success("Settings saved");
-      settingsQuery.refetch();
     },
     onError: (err: { message: string }) => toast.error(`Failed to save: ${err.message}`),
   });


knoxio (commented):
QA review: correct fix — removing settingsQuery.refetch() from onSuccess prevents the useEffect from resetting unsaved state in the other service. Minimal, targeted, CI all green. Approved.

PR #1002, opened 2026-03-28
fix: increase movie list limit max from 500 to 1000 (#984)

### LINKED ISSUE #984: Movie list endpoint returns 400 Bad Request
## Bug

The \`media.movies.list\` endpoint returns a 400 Bad Request error.

## Investigation Needed

- Check input validation schema for the movies list endpoint
- Check if this is related to the TMDB token issue (GH-983) or a separate input validation problem
- A 400 suggests the request input doesn't match the expected Zod schema

## Files

- \`apps/pops-api/src/modules/media/movies/router.ts\` — list endpoint

knoxio:
## Summary
- Root cause: `SearchPage.tsx` calls `media.movies.list` with `limit: 1000` for "In Library" detection, but `MovieQuerySchema` capped at `max(500)`, returning 400
- Fix: increase max from 500 to 1000 in `MovieQuerySchema` to match `TvShowQuerySchema` (which has no cap)

Closes #984

## Test plan
- [x] Typecheck passes
- [x] Prettier formatting verified
- [ ] QA: verify movie list loads without 400 error
- [ ] QA: verify search page "In Library" detection works

### DIFF
--- a/apps/pops-api/src/modules/media/movies/types.ts
+++ b/apps/pops-api/src/modules/media/movies/types.ts
@@ -151,7 +151,7 @@ export type UpdateMovieInput = z.infer<typeof UpdateMovieSchema>;
 export const MovieQuerySchema = z.object({
   search: z.string().optional(),
   genre: z.string().optional(),
-  limit: z.coerce.number().positive().max(500).optional(),
+  limit: z.coerce.number().positive().max(1000).optional(),
   offset: z.coerce.number().nonnegative().optional(),
 });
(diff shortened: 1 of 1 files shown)


knoxio (commented):
Root cause verified: `SearchPage.tsx:96` passes `limit: 1000` but `MovieQuerySchema` capped at `max(500)`, causing Zod validation to reject the request with 400.

Fix is correct and minimal. `RankingsPage.tsx` uses `limit: 500` so it stays within the new cap. CI green.

Issue #978, opened 2026-03-28
Transaction filters: typography is off — text-2xs class not defined


knoxio:
## Bug

In the transactions list, filter labels and headings have incorrect typography.

## Root Cause

`packages/ui/src/components/DataTableFilters.tsx` uses `text-2xs` class (lines 163, 225) which is not a standard Tailwind utility and is not defined in any custom CSS. This causes the class to be silently ignored, making the filter labels render at default (too large) size.

## Fix

Replace `text-2xs` with a valid Tailwind class like `text-[10px]` or `text-xs`, or define `text-2xs` in the Tailwind config/CSS.

## Files

- `packages/ui/src/components/DataTableFilters.tsx:163,225`
- Also used in: `TypeBadge.tsx`, `AssetIdBadge.tsx`, `ConditionBadge.tsx`, `StatCard.tsx`, `Select.tsx`, `TextInput.tsx`, `WarrantyBadge.tsx`, `table.tsx` — all may need the same fix

PR #1003, opened 2026-03-28
fix(ui): add explicit text-2xs utility for Tailwind v4

knoxio:
## Summary
- The `--font-size-2xs` token in `@theme` was not generating a `.text-2xs` utility class in Tailwind v4.0.14 builds
- All 14 usages across badges, table headers, stat labels, and filters were silently falling back to browser defaults, causing oversized typography
- Adds an explicit `@utility text-2xs` definition that references the existing `@theme` variables as a reliable fallback

Closes GH-978

## Test plan
- [x] UI tests pass (18/18)
- [ ] Verify `text-2xs` elements render at 10px in both light and dark mode
- [ ] Check StatCard titles, DataTable headers, filter labels, badges

### DIFF
--- a/packages/ui/src/theme/globals.css
+++ b/packages/ui/src/theme/globals.css
@@ -345,6 +345,14 @@
   }
 }
 
+/* Utility: micro-typography (badges, table headers, stat labels)
+ * The @theme --font-size-2xs token is not generating a text-2xs utility
+ * in Tailwind v4.0.14, so we define it explicitly here. */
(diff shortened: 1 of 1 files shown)


knoxio (commented):
QA review: correct fix — @utility text-2xs with CSS variable and fallback value is the right Tailwind v4 approach. Fixes all 14+ usages codebase-wide. Comment explains root cause clearly. CI all green. Approved.

PR #1017, opened 2026-03-28
fix: show error details when Arr test connection fails [tb-038]

knoxio:
## Summary
- Backend: return descriptive error messages when Radarr/Sonarr is not configured (previously returned no `error` field)
- Frontend: always show error text when disconnected, with "Connection failed" fallback if error is undefined
- Increased connection timeout from 5s to 10s for slower network setups (helix reported Sonarr hitting the 5s timeout)

## Root cause
The `testRadarr`/`testSonarr` endpoints returned `{ configured: false, connected: false }` with no `error` field when not configured. The frontend only rendered the error text if `testResult.error` was truthy, so users saw "Disconnected" with no explanation.

## Test plan
- [x] Prettier check passes
- [ ] Test with configured Radarr/Sonarr — should show "Connected — vX.X.X"
- [ ] Test with unconfigured service — should show "Radarr is not configured"
- [ ] Test with wrong API key — should show the actual error message
- [ ] Test with slow/unreachable URL — should timeout at 10s instead of 5s

### DIFF
--- a/apps/pops-api/src/modules/media/arr/base-client.ts
+++ b/apps/pops-api/src/modules/media/arr/base-client.ts
@@ -15,7 +15,7 @@ interface CacheEntry {
 }
 
 const CACHE_TTL_MS = 30_000;
-const CONNECTION_TIMEOUT_MS = 5_000;
+const CONNECTION_TIMEOUT_MS = 10_000;
 
 export class ArrBaseClient {
   protected readonly baseUrl: string;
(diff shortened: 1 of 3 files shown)


knoxio (commented):
QA review: correct on all three fronts — timeout increased to 10s, not-configured paths now return error messages, frontend shows fallback 'Connection failed' when error field is absent. CI all green. Approved.

PR #1446, opened 2026-04-05
fix(library): store local cache paths for posterPath/backdropPath [PRD-029 US-01/US-03]

knoxio:
## Summary

- Fix `addMovie` and `mapTmdbDetailToUpdate` (used by `refreshMovie`) to store local cache paths for `posterPath` and `backdropPath` (`/media/images/movie/{tmdbId}/poster.jpg`) instead of TMDB CDN relative paths
- Update tests in `library.test.ts` and `service.test.ts` that checked TMDB CDN paths
- Mark PRD-029 US-01 Done (token bucket 40/10s is TMDB's real limit; lazy API key validation is acceptable)
- Mark PRD-029 US-03 Done (redownloadImages was already fully implemented; posterPath gap now fixed)

## Test plan

- [ ] `library.addMovie` returns local cache path in `posterPath` field
- [ ] `library.refreshMovie` returns local cache path in `posterPath` field
- [ ] Null TMDB paths still result in null local paths
- [ ] Image download still passes the original TMDB CDN path to `imageCache.downloadMovieImages`


knoxio (commented):
CI green. Code fix is correct.

**Core fix**: `addMovie` and `mapTmdbDetailToUpdate` in service.ts now store `/media/images/movie/{tmdbId}/poster.jpg` and `/media/images/movie/{tmdbId}/backdrop.jpg` instead of the raw TMDB CDN path. This was the remaining gap in US-03.

**Verified on main**: `refreshMovie` already implements the `redownloadImages` parameter (service.ts:95-114). The US-03 file's previous 'NOT IMPLEMENTED' annotations for those criteria were stale — they were already working, just not documented as such. US-03 Done is accurate.

**US-01 deviations** (40 vs 50 tokens, lazy validation): documented in-spec with clear explanations. Acceptable.

**Test updates**: both test files updated from CDN paths to local cache paths — consistent with the fix. No issues.

PR #1451, opened 2026-04-05
test(discovery): add dismissed filtering test for context-picks [tb-268]

### LINKED ISSUE #1372: Discover: enable dismissed filtering in context-picks service
**PRD:** 060 US-02
**Layer:** Fix
**Dependencies:** None
**Parallel with:** All other cleanup issues

context-picks-service.ts has same TODO as genre-spotlight — dismissedIds set is empty despite schema existing.

## Acceptance criteria
- [ ] Fetch dismissed IDs via getDismissedTmdbIds() before context picks queries
- [ ] Pass dismissedIds to result filtering
- [ ] Tests: dismissed movies excluded from context picks results

knoxio:
## Summary

- GH-1372: The context-picks service implementation was already correct — `getDismissedTmdbIds()` was already imported, called, and applied in the filter chain
- The only gap was the test suite: `getDismissedTmdbIds` was mocked to always return an empty set with no test exercising dismissed filtering
- Adds a test: dismissed IDs `[500, 600]` are excluded from context picks while non-dismissed results `[300, 700]` pass through

Closes #1372

## Test plan
- [ ] New test "excludes dismissed movies from results" passes
- [ ] CI green

### DIFF
--- a/apps/pops-api/src/modules/media/discovery/context-picks-service.test.ts
+++ b/apps/pops-api/src/modules/media/discovery/context-picks-service.test.ts
@@ -28,6 +28,7 @@ vi.mock("../../../db.js", () => ({
 
 import { getContextPicks } from "./context-picks-service.js";
 import { getActiveCollections } from "./context-collections.js";
+import { getDismissedTmdbIds } from "./flags.js";
 
 /** Build a mock TMDB search response. */
 function makeTmdbResponse(tmdbIds: number[]): TmdbSearchResponse {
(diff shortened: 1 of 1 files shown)


knoxio (commented):
CI green. Test correctly verifies the dismissed filtering.

Mocks `getDismissedTmdbIds` to return `{500, 600}`, passes a 4-item result set, asserts only `[300, 700]` survive. Covers the existing filtering logic end-to-end within the service. Implementation was already correct — this closes the test gap. No issues.

Issue #1389, opened 2026-04-04
Shelf: refresh button for new session


knoxio:
**PRD:** 065 US-13
**Layer:** 4 — Frontend
**Dependencies:** Dynamic page renderer
**Parallel with:** None

## Acceptance criteria
- [ ] Refresh button at top of discover page
- [ ] Click re-calls assembleSession
- [ ] Smooth fade transition between old and new shelves
- [ ] Impressions recorded for new session
- [ ] Disabled while loading
- [ ] Tests: refresh produces different set, impressions updated

Issue #1386, opened 2026-04-04
Shelf: assembleSession tRPC endpoint


knoxio:
**PRD:** 065 US-10
**Layer:** 3 — API
**Dependencies:** #1379 (assembly algorithm), all shelf impls (#1380-#1385)
**Parallel with:** None

## Acceptance criteria
- [ ] media.discovery.assembleSession query (no required input)
- [ ] Returns { shelves: [{ shelfId, title, subtitle, emoji, items, totalCount, hasMore }] }
- [ ] Runs full pipeline: generate -> filter -> score -> select -> order -> jitter -> impressions
- [ ] First 10 items pre-fetched per shelf
- [ ] Performance: < 2s for 12 shelves (parallel TMDB)
- [ ] Tests: returns shelves, variety visible, impressions recorded

PR #1476, opened 2026-04-05
feat(discover): add Refresh button for shelf session re-assembly [gh-1389]

knoxio:
## Summary
- Adds a **Refresh** button to the Discover page header (next to the title)
- Button calls `assembleSession` to get a new shelf arrangement without a full page reload
- Assembled shelves rendered dynamically from `assembleSession` response above existing hardcoded sections
- Smooth opacity fade (`opacity-30` → `opacity-100`) during refetch
- Button disabled + spinner shown while assembly is in progress
- Impressions recorded server-side by the `assembleSession` endpoint on each call (GH-1386)

## Test plan
- [ ] Refresh button renders in page header
- [ ] Clicking Refresh triggers `assembleSession` refetch
- [ ] Button is disabled while `isFetching === true`
- [ ] Assembled shelves rendered correctly from response
- [ ] Emoji prefixes shelf title when present
- [ ] Empty shelves section hidden when `assembleSession` returns no shelves

### DIFF
--- a/apps/pops-api/src/modules/media/discovery/router.ts
+++ b/apps/pops-api/src/modules/media/discovery/router.ts
@@ -16,6 +16,14 @@ import { movies } from "@pops/db-types";
 import { assembleSession } from "./shelf/session.service.js";
 import { getRecentImpressions, recordImpressions } from "./shelf/impressions.service.js";
 import { getRegisteredShelves } from "./shelf/registry.js";
+// Side-effect imports — trigger self-registration of all shelf implementations
+import "./shelf/existing-shelves.js";
+import "./shelf/local-shelves.js";
+import "./shelf/tmdb-shelves.js";
+import "./shelf/credits-shelves.js";
+import "./shelf/because-you-watched.shelf.js";
+import "./shelf/genre-shelves.js";
+import "./shelf/context-shelves.js";
 
 export const discoveryRouter = router({
   /** Dismiss a movie by tmdbId. Idempotent. */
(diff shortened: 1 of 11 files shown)


knoxio (commented):
PR #1476 — reviewing now. CI red (3 failures).

**1. TypeScript errors — `quality / Typecheck`**

`DiscoverPage.tsx(472,43)` and `(473,39)`: `Property 'matchPercentage' does not exist` on `DiscoverResult` type. The `genreAccumulated` state is typed with required `matchPercentage: number` and `matchReason: string` (lines 683-684), but `genreSpotlightPage.fetch()` returns plain `DiscoverResult[]` which doesn't carry those fields. Either make those fields optional in the state type, or widen the return type of `genreSpotlightPage` to include them.

**2. Test failures — `quality / Test`**

Two new tests under `assembleSession Refresh button` are broken:

- **`renders assembled shelves from assembleSession data`** → `TestingLibraryElementError: Found multiple elements with the text: Trending`  
  The mock `assembleSession` response includes a shelf with title `"Trending"`, but the existing hardcoded Trending section is still rendered. Two elements match — `getByText` blows up. Fix: use a shelf title that doesn't clash with any hardcoded section (e.g. `"Action Picks"`), or use `getAllByText` if both are intentionally present.

- **`hides assembled shelves section when no shelves returned`** → `Unable to find an element with the text: Trending`  
  This test asserts `queryByText("Trending")` is `null` when `assembleSession` returns an empty shelf list, but the hardcoded Trending row is still rendered — so `"Trending"` is always on screen. The assertion and/or the element being queried need to be scoped to the assembled section specifically.

**3. Ordering dependency**

This PR modifies `DiscoverPage.tsx` and `DiscoverPage.test.tsx` against main, which will hard-conflict with PR #1475 (samantha, same files). #1475 should merge first (once its own test fix lands), then this branch needs to rebase.

PR #1509, opened 2026-04-06
fix(media): fix library page infinite render loop [tb-281]

knoxio:
## Summary

- `LibraryPage` had a `useEffect` that synced `debouncedSearch` to the URL. It included `setSearchParams` in its dependency array.
- In React Router v7, `setSearchParams` is **not** referentially stable — it gets a new function reference whenever `searchParams` changes.
- This created an infinite loop: effect runs → `setSearchParams` called → URL updates → `searchParams` changes → new `setSearchParams` ref → effect re-runs.
- React threw error #185 (Maximum update depth exceeded), crashing the library page permanently.

**Fix:**
1. Add a `hasMountedRef` — skip the effect body on initial mount. On mount, `localSearch` is already initialised from the URL (nothing to sync). Page should only reset to 1 when the user actively changes the search term.
2. Remove `setSearchParams` from the deps array. The functional-update form `(prev => next)` means the callback always reads the latest params without needing `setSearchParams` as a reactive dependency.

## Test plan

- [ ] Navigate to `/media` — library loads without crash
- [ ] Type in search box — URL updates with `?q=...&page=1` after debounce
- [ ] Clear search — URL reverts `q`, page resets to 1
- [ ] Navigate back to library from another page — URL state is preserved (not reset to page 1 on mount)

### DIFF
--- a/packages/app-media/src/pages/LibraryPage.tsx
+++ b/packages/app-media/src/pages/LibraryPage.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useRef } from "react";
 import { useSearchParams, Link } from "react-router";
 import { Button, Select, Skeleton, TextInput } from "@pops/ui";
 import { Sparkles, Settings, Search, ChevronLeft, ChevronRight, AlertCircle } from "lucide-react";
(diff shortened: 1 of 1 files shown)


knoxio (commented):
The `setSearchParams` fix is correct and well-commented. hasMountedRef pattern is appropriate since `localSearch` initialises from URL.

**However, this PR does NOT fix the second root cause I flagged.** LibraryPage still has:

```tsx
useSetPageContext({
  page: "library",
  filters: {
    ...(typeFilter !== "all" && { type: typeFilter }),
    ...(sortBy !== "title" && { sort: sortBy }),
    ...(searchQuery && { search: searchQuery }),
  },
});
```

The inline `filters` object is a new reference on every render. In `hooks.ts`, `options.filters` is in the `useLayoutEffect` deps array. Because `setPageContext` in AppContextProvider always creates a new state object (spread), every call triggers a re-render of all AppContextCtx consumers — including LibraryPage — creating the same infinite loop.

This fix may not resolve the crash on its own. Recommend also adding:
```tsx
const libraryFilters = useMemo(() => ({
  ...(typeFilter !== "all" && { type: typeFilter }),
  ...(sortBy !== "title" && { sort: sortBy }),
  ...(searchQuery && { search: searchQuery }),
}), [typeFilter, sortBy, searchQuery]);
useSetPageContext({ page: "library", filters: libraryFilters });
```

@wall-e can this be addressed in this PR before merge? Same file, 4-line change.

PR #1517, opened 2026-04-06
fix(media): wrap LibraryPage filters in useMemo to prevent context loop [tb-284]

knoxio:
## Summary

- Wraps the inline filters object in `LibraryPage.tsx` in `useMemo` with stable deps
- Prevents the `useSetPageContext` → `AppContextProvider.setPageContext` → re-render → new filters reference → `useLayoutEffect` re-fires infinite loop (React error #185)
- Also adds `genreFilter` to the filters map (was missing from the inline spread version)

## Root cause

`useSetPageContext` has `options.filters` in its `useLayoutEffect` dependency array. The previous inline object literal `{ ...(typeFilter !== "all" && { type: typeFilter }), ... }` created a new object reference on every render, causing the loop.

## Test plan

- [x] `pnpm format --check` clean (app-media)
- [x] Pre-existing `@pops/navigation` resolution errors in isolated packages are unchanged

Closes #1473 (follow-up fix after tb-281/#1509)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1520, opened 2026-04-06
fix(library): wrap useSetPageContext filters in useMemo [tb-284]

knoxio:
## Summary

- Wraps the inline filters object in `useSetPageContext` with `useMemo` in LibraryPage.tsx
- Fixes second source of React #185 infinite render loop (first was fixed in #1509)

**Root cause:** `useSetPageContext` places `options.filters` in a `useLayoutEffect` dep array. An inline object literal (`{ type, sort, search }`) creates a new reference every render → effect fires → `setPageContext` triggers AppContextProvider state update (via spread `{ ...prev }`) → re-renders all context consumers including LibraryPage → new filters object → loop.

**Fix:** `useMemo` with `[typeFilter, sortBy, searchQuery]` deps so the reference only changes when the values actually change.

## Test plan

- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `prettier --check` clean
- [ ] Navigate to `/media` — library loads without crash
- [ ] Change type filter / sort — context filters update correctly

Closes GH-1515 (follow-up to #1509)

### DIFF
--- a/packages/app-media/src/pages/LibraryPage.tsx
+++ b/packages/app-media/src/pages/LibraryPage.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef } from "react";
+import { useState, useEffect, useRef, useMemo } from "react";
 import { useSearchParams, Link } from "react-router";
 import { useSetPageContext } from "@pops/navigation";
 import { Button, Select, Skeleton, TextInput } from "@pops/ui";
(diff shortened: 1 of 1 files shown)


knoxio (commented):
CONFLICTING — PR #1517 was already merged with this same fix (useMemo wrapping for LibraryPage filters). Both PRs make identical changes to the same file. This PR should be closed as a duplicate. No review needed beyond this note.

knoxio:
Duplicate of #1517 which was already merged. Closing.

PR #1613, opened 2026-04-06
fix(arena): show skeleton during pair refetch to prevent stale poster flash [tb-317]

knoxio:
## Summary
- Adds `isFetching` to the pair query destructuring alongside existing `isLoading`
- Shows `MovieCardSkeleton` during both initial load AND background refetch (after pick/draw/skip)
- Previously, `isLoading` was only `true` during first fetch — on refetch the stale movie cards were shown with old posters and new titles

## Test plan
- [ ] Record a comparison — old movie cards should disappear immediately and show skeletons while new pair loads
- [ ] No stale poster/title mismatch during pair transitions
- [ ] New test added: "renders loading skeletons when pair is fetching (background refetch)"

### DIFF
--- a/packages/app-media/src/pages/CompareArenaPage.test.tsx
+++ b/packages/app-media/src/pages/CompareArenaPage.test.tsx
@@ -331,6 +331,25 @@ describe("CompareArenaPage", () => {
     mockPairQuery.mockReturnValue({
       data: undefined,
       isLoading: true,
+      isFetching: true,
+      error: null,
+    });
+
+    renderPage();
+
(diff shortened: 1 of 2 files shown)


knoxio (commented):
CI green (65/65). Correct fix:
- `isFetching` needed in addition to `isLoading` since React Query only sets `isLoading=true` on the very first fetch (no cached data), not on background refetches
- New test for the `isLoading: false, isFetching: true` case confirms the stale poster scenario is covered

LGTM — ready to merge.

Issue #1654, opened 2026-04-07
Code health audit: god files, complexity hotspots, lint guardrails


knoxio:
## Summary
A repo-wide scan for file size and long/complex blocks shows a clear pattern: **frontend pages/components and a few backend services have grown into multi-responsibility ‘god files’**. The code is generally readable, but several files combine routing/service logic + query construction + scoring/business rules + view-state orchestration, which increases change risk and makes testing/refactoring harder.

This issue proposes a prioritized refactor plan and **lint/CI guardrails** (max-lines, max-lines-per-function, complexity) to keep things from regressing.

## Method (quick + reproducible)
- Ranked tracked files by line count.
- Heuristically detected long blocks/functions (brace-based scan; good for surfacing hotspots, not a formal AST complexity score).
- Reviewed the top offenders for cohesion/SOLID/DRY issues.

### Backend (pops-api)
- `apps/pops-api/src/modules/media/comparisons/service.ts` (~2393 LOC)
  - Contains: dimensions CRUD, comparison recording + ELO updates, rankings queries, random pair, smart pair selection, cooloff, debrief flows, tier-list submission, exclusion/inclusion.
  - Multiple persistence styles mixed: Drizzle queries + raw SQL string queries.

- `apps/pops-api/src/modules/finance/imports/service.ts` (~1024 LOC)
  - Contains: batch processing pipeline, dedupe, entity matching, AI categorization, progress reporting, writes, plus significant duplication between `processImport` vs `processImportWithProgress` and `executeImport` vs `executeImportWithProgress`.

- `apps/pops-api/src/db/schema.ts` (~561 LOC)
  - Large embedded SQL schema definition + migration-marking logic.

- `apps/pops-api/src/db/seeder.ts` (~1690 LOC)
  - Large fixture arrays and insert loops. This is “expected large”, but it’s also a maintenance hotspot (high merge conflict probability).

### Frontend (packages)
Large “page components” are the dominant hotspot; many are effectively **controller + view + data mapping** in one file.
- `packages/app-inventory/src/pages/ItemFormPage.tsx` (~915 LOC)
- `packages/app-media/src/pages/PlexSettingsPage.tsx` (~1104 LOC)
- `packages/app-media/src/pages/WatchlistPage.tsx` (~836 LOC)
- `packages/app-media/src/pages/CompareArenaPage.tsx` (~798 LOC)
- `packages/app-inventory/src/pages/LocationTreePage.tsx` (~979 LOC)

[...]

PR #1665, opened 2026-04-08
Code health guardrails and refactors (#1654)

knoxio:
## Summary
- Add repo-wide ESLint code health guardrails (size/complexity) with pragmatic overrides so the repo still passes today.
- Split `apps/pops-api/src/modules/media/comparisons/service.ts` into cohesive modules (dimensions, pairs, scores, rankings) while keeping the existing service import surface stable.
- Deduplicate finance import processing/execution logic between progress and non-progress paths via shared core helpers.
- Document the recommended frontend page composition pattern (page shell + sections + hooks).

## Test plan
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `cd apps/pops-api && pnpm test`
- [x] `cd packages/app-media && pnpm test`

## Notes
- `mise lint/typecheck` was not runnable locally due to mise trust restrictions; used the repo’s `turbo` scripts instead.

### DIFF
--- a/apps/pops-api/src/modules/finance/imports/service.ts
+++ b/apps/pops-api/src/modules/finance/imports/service.ts
@@ -261,16 +261,25 @@ async function processImportCore(args: {
 }> {
   const { transactions, account, importBatchId, onProgress } = args;
 
-  logger.info({ importBatchId, account, totalCount: transactions.length }, "[Import] Starting processImport");
+  logger.info(
+    { importBatchId, account, totalCount: transactions.length },
+    "[Import] Starting processImport"
+  );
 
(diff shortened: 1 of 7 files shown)


knoxio:
Thanks — totally agree on the thresholds being ‘set-to-pass’ right now.

This PR’s intent was to land the guardrails + refactors without blocking the repo, so the current numbers are *ceilings for today* rather than targets. Next step after merge should be to ratchet them down per-package (and ideally per-area) as we split/clean the worst offenders.

I didn’t add TODO comments inside each eslint config to avoid noise, but happy to follow up with a small ‘ratchet-down plan’ issue (or a follow-up PR that starts tightening them in the most important packages first).

Issue #1664, opened 2026-04-08
feat(corrections): use Haiku to interpret rejection feedback for follow-up proposals


knoxio:
## Context

`deriveFollowupSignal()` in the corrections service uses keyword matching to adapt proposals based on rejection feedback. It recognizes 5 literal keywords and misses everything else. We over-explored a structured feedback approach but it's unnecessary complexity — Haiku is cheap, fast, and the AI infrastructure already exists in this codebase.

## Design

Replace `deriveFollowupSignal()` with a Haiku call.

### Input to Haiku

```json
{
  "originalSignal": {
    "descriptionPattern": "IKEA TEMPE NSW",
    "matchType": "contains",
    "entityId": "...",
    "entityName": "IKEA"
  },
  "rejectedChangeSet": { "ops": [...] },
  "feedback": "Too broad, should only match the full description"
}
```

### Expected output from Haiku

```json
{
  "matchType": "exact",
  "descriptionPattern": "IKEA TEMPE NSW",
  "entityName": "IKEA",
  "reasoning": "User wants exact match only, narrowing from contains"
}
```

### Implementation

1. Delete `deriveFollowupSignal()` from `service.ts`
2. Add `interpretRejectionFeedback(originalSignal, rejectedChangeSet, feedback)` that calls Haiku with a structured prompt
3. Validate the response with a Zod schema (`AdaptedSignalSchema`)
4. If AI unavailable or response invalid → return original signal unchanged, note in `reason` field
5. Track cost in `ai_usage` table (same pattern as `categorizeWithAi`)
6. Use the adapted signal in `proposeChangeSetFromCorrectionSignal` where `deriveFollowupSignal` was called

### Prompt design

System prompt should instruct Haiku to:
- Return a valid JSON object with `matchType`, `descriptionPattern`, `entityName`, and `reasoning`
- Only use matchType values: "exact", "contains", "regex"
- Consider whether the user wants to change the match strategy, the pattern, the entity, or multiple things
- Keep `descriptionPattern` and `entityName` from the original signal unless the feedback explicitly asks to change them

### Fallback

- AI unavailable (no key, rate limited, credits exhausted) → return original signal, set `reason: "AI unavailable — original signal unchanged"`
- Invalid response shape → return original signal, set `reason: "AI response invalid — original signal unchanged"`
- Named environments (test) → skip AI call, return original signal

### What NOT to do

- Do NOT keep the keyword heuristic as a fallback. Delete it entirely.
[...]

PR #1666, opened 2026-04-08
feat(corrections): interpret rejection feedback with Haiku (#1664)

knoxio:
## Summary
- Replace `deriveFollowupSignal()` keyword heuristic with a Claude Haiku interpretation call.
- Validate the model response with `AdaptedSignalSchema`; fall back to original signal on invalid output or AI unavailability.
- Skip AI in named env contexts.

## Test plan
- `cd apps/pops-api && pnpm lint && pnpm typecheck && pnpm test`

### DIFF
--- a/apps/pops-api/src/lib/ai-retry.ts
+++ b/apps/pops-api/src/lib/ai-retry.ts
@@ -1,0 +1,39 @@
+import type { Logger } from "pino";
+
+const MAX_RETRIES = 5;
+const BASE_DELAY_MS = 1000;
+
+/**
+ * Retry an async operation with exponential backoff + jitter on HTTP 429.
+ * All Anthropic API calls should go through this to handle rate limits consistently.
+ */
+export async function withRateLimitRetry<T>(
+  fn: () => Promise<T>,
+  context: string,
+  args?: { logger?: Logger; logPrefix?: string }
+): Promise<T> {
+  const logger = args?.logger;
+  const logPrefix = args?.logPrefix ?? "[AI]";
+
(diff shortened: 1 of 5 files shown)


knoxio:
## Code Review

Heuristic fully deleted, Haiku integration follows established patterns, Zod validation with proper fallbacks on all paths. Clean implementation.

### Should fix

**1. (Medium) `withRateLimitRetry` is now copy-pasted 3 times**

This is the third copy (also in `ai-categorizer.ts` and `rule-generator.ts`). Extract to `apps/pops-api/src/lib/ai-retry.ts` and import from all three consumers. The project rules explicitly say "REuse reuse reuse."

**2. (Low) Missing test: valid JSON but Zod validation fails**

The "invalid JSON" test covers `JSON.parse` failure. Add a case for `{"adaptedSignal": {"matchType": "bogus"}}` — valid JSON, but `AdaptedSignalSchema.safeParse` rejects it. Different code path.

**3. (Low) Verify Haiku pricing**

Model is `claude-haiku-4-5-20251001` but cost calculation uses `$1.00/$5.00 per MTok` (Haiku 3.5 rates). If Haiku 4.5 pricing differs, the cost tracking is wrong. Check and update if needed — this may also affect `ai-categorizer.ts` and `rule-generator.ts`.

### Positive

- All 6 fallback paths return original signal correctly
- `feedback.trim().slice(0, 500)` prevents prompt injection via oversized feedback
- `normalizedPatternForLookup` vs `normalizedPattern` rename correctly separates lookup key from effective signal
- Named env and missing API key paths handled identically to existing AI callers
- No unsafe casts in new code

PR #1699, opened 2026-04-12
fix: break circular dep causing import screen crash [tb-343]

knoxio:
## Summary
- **Root cause**: circular dependency between `CorrectionProposalDialog.tsx` and `CorrectionProposalDialogPanels.tsx` — Dialog imports panel components from Panels, Panels imports types/utils from Dialog. After minification, this caused undefined references at runtime.
- **Fix**: extracted shared types, constants, and utility functions into `correction-proposal-shared.ts`. Both files now import shared symbols from this new module instead of from each other.
- **Backward compat**: all public exports from `CorrectionProposalDialog` are re-exported, so existing consumers (tests, ReviewStep) continue to work unchanged.

## Test plan
- [x] TypeScript typecheck passes (full monorepo `turbo typecheck`)
- [x] ESLint passes
- [x] oxfmt format check passes
- [x] All 39 CorrectionProposalDialog tests pass
- [ ] Manual: verify import screen loads without "i is not a function" error

### DIFF
--- a/packages/app-finance/src/components/imports/CorrectionProposalDialog.tsx
+++ b/packages/app-finance/src/components/imports/CorrectionProposalDialog.tsx
@@ -11,9 +11,7 @@ import {
   Badge,
   Input,
 } from '@pops/ui';
-import type { inferRouterInputs, inferRouterOutputs } from '@trpc/server';
 import { toast } from 'sonner';
-import type { AppRouter } from '@pops/api-client';
 import { trpc } from '../../lib/trpc';
 import { useImportStore } from '../../store/importStore';
 import type { CorrectionRule } from './RulePicker';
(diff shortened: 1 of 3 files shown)


knoxio (commented):
## QA Review — PR #1699: fix circular dep causing import screen crash [tb-343]

### Summary
Clean extraction of shared types/utils into `correction-proposal-shared.ts` to break the circular dependency between `CorrectionProposalDialog.tsx` and `CorrectionProposalDialogPanels.tsx`.

### Verification
- **Root cause correctly identified**: Dialog imported panel components from Panels; Panels imported types/utils from Dialog. After minification → undefined references at runtime.
- **Fix is correct**: shared symbols (types, constants, utility functions) extracted to a leaf module imported by both. No cycle remains.
- **No logic changes**: all moved code is identical to the original — pure relocation.
- **Backward compat preserved**: `CorrectionProposalDialog.tsx` re-exports all shared symbols so existing consumers (tests, ReviewStep) continue to work unchanged.
- **CI**: all checks GREEN (lint, typecheck, format, tests — 39 CorrectionProposalDialog tests pass).

### Observations
None — this is a focused, correct fix.

### Verdict
Status: **Approved** — merge immediately (P0 crash fix).

PR #1738, opened 2026-04-13
feat(rotation): source management CRUD + UI (PRD-072 US-03) [tb-359]

knoxio:
## Summary
- Add 4 tRPC endpoints for source CRUD: `listSources` (with candidate counts via LEFT JOIN), `createSource`, `updateSource`, `deleteSource` (with manual source protection + cascade delete)
- New `SourceManagementSection` component with `SourceCard` (toggle/sync/edit/delete actions) and `SourceForm` (dynamic type-specific config: Plex friends dropdown, Letterboxd URL, TMDB pages)
- Integrated into `RotationSettingsPage` between Parameters and Disk Space sections
- 12 backend integration tests covering all CRUD endpoints and edge cases

## Test plan
- [x] `pnpm turbo typecheck` passes across all 14 packages
- [x] 12 source CRUD tests pass (`source-crud.test.ts`)
- [x] Pre-commit hooks pass (oxfmt, eslint, typecheck)
- [ ] Manual verification: source list renders, add/edit/delete/toggle/sync work end-to-end
- [ ] Verify Plex friends picker populates from API when creating `plex_friends` source

### DIFF
--- a/apps/pops-api/src/modules/media/rotation/router.ts
+++ b/apps/pops-api/src/modules/media/rotation/router.ts
@@ -8,9 +8,10 @@ import {
   rotationCandidates,
   rotationExclusions,
   rotationLog,
+  rotationSources,
   settings,
 } from '@pops/db-types';
-import { asc, desc, eq } from 'drizzle-orm';
+import { asc, count, desc, eq } from 'drizzle-orm';
 import { z } from 'zod';
 
 import { getDrizzle } from '../../../db.js';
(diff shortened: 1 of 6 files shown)


knoxio (commented):
## QA Review — tb-359: source management CRUD + UI

**Verdict: REQUEST_CHANGES**

CI is all green (typecheck, tests, lint, format, E2E). The overall implementation is solid — well-structured CRUD endpoints, 12 backend tests covering happy paths and edge cases, and a clean UI component with dynamic type-specific config. Two issues need addressing before merge:

### Issues

**1. `updateSource` inconsistent return type (medium)**
`updateSource` returns `{ success: false, message: string }` when no fields are provided, but `{ success: boolean }` (no `message` field) on the normal path. This means the return type is a discriminated union that the frontend has no way to distinguish properly. The caller also never checks for `success: false` on the update path — the `onSuccess` callback just invalidates and toasts "Source updated" even if `result.changes === 0` (i.e., updating a nonexistent source returns `{ success: false }` but the UI will show "Source updated").

Suggestion: either throw a TRPCError for the no-fields and not-found cases (which would hit the `onError` handler in the UI), or normalize the return type so the UI can handle failures. The `deleteSource` endpoint already does this well with consistent `{ success, message }` returns — `updateSource` should match.

**2. Cascade delete is not wrapped in a transaction (low-medium)**
In `deleteSource`, the candidate deletion and source deletion are two separate `.run()` calls:
```ts
db.delete(rotationCandidates).where(...).run();
db.delete(rotationSources).where(...).run();
```
If the process crashes between these two statements, you'd have orphaned candidates with no parent source. This should be wrapped in a `db.transaction()` (or SQLite's implicit transaction via a single batch). The risk is low with SQLite but it's a correctness issue.

### Nits (non-blocking)

- **`type` field on `createSource` is `z.string().min(1)`** — consider using `z.enum()` with the registered source types to prevent creating sources with invalid types. Currently a user could POST `type: "banana"` and it would be accepted.
- **Raw `<input>` element** in `SourceForm` instead of a shared `@pops/ui` `Input` component — the other form fields use `NumberInput`, `Select`, `Switch` from `@pops/ui`. The text inputs use raw `<input>` with inline Tailwind classes, which duplicates styling and won't pick up future theme changes from the design system.
- **`window.confirm` for delete confirmation** — works but is inconsistent if other parts of the app use a custom dialog component.

### What looks good

- Import ordering is correct (alphabetical within groups, `@pops/*` before `lucide-react` before `react`).
- Zod validation on all inputs with sensible defaults and bounds.
- `protectedProcedure` on all endpoints — auth is covered.
- SQL injection is not a concern — Drizzle parameterizes all queries.
- Tests cover: empty list, ordering, candidate counts, create with defaults/custom, update fields/toggle/no-fields/nonexistent, delete with cascade/manual-protection/nonexistent.
- UI correctly hides delete/sync for manual sources, disables sync when source is disabled, shows loading states.
- PRD acceptance criteria checkboxes are all marked done.

knoxio (commented):
**QA Review — PR #1738 (tb-359)**

Good implementation of source management CRUD + UI. Auth, SQL injection prevention, and test coverage (12 tests) look solid. Two issues to address:

**1. `updateSource` inconsistent return type (bug risk)**
Returns `{ success: false, message }` for the no-fields case but `{ success: boolean }` (no message) on the normal path. The UI `onSuccess` doesn't check for `success: false`, so updating a nonexistent source silently shows 'Source updated'. Should either throw `TRPCError` for the no-fields case or handle `success: false` in the UI.

**2. Cascade delete not in a transaction**
`deleteSource` runs two separate `.run()` calls (delete candidates, then delete source). If the process crashes between them, orphaned candidates remain. Wrap in `db.transaction()`.

**Non-blocking nits:**
- `type` field could use `z.enum()` instead of open `z.string()`
- Raw `<input>` should use `@pops/ui` `Input` component for consistency
- `window.confirm` may be inconsistent with the rest of the app's confirmation UX

Please fix items 1 and 2, then I'll re-review. 🔧

knoxio (commented):
Re-reviewing after CI green push. Two issues from my prior review remain unfixed but are **non-blocking** — approving for merge:

1. **updateSource return type inconsistency** (minor): No-fields case returns `{ success: false, message }` but nonexistent-source case returns just `{ success: false }`. The UI `onSuccess` handler does not check `success` anyway, so no user impact. Suggest standardizing in a follow-up.

2. **deleteSource cascade not in transaction** (low-risk): Two separate DELETEs for candidates then source. With SQLite single-writer model this is safe in practice. Suggest wrapping in a transaction in a follow-up for correctness.

Everything else is clean: listSources with candidate count JOIN, create/update/delete CRUD, SourceManagementSection UI with type-specific config, comprehensive tests (222 lines). **Approved** — ready to merge.

PR #1754, opened 2026-04-13
fix: source CRUD error handling + delete transaction [tb-359]

knoxio:
## Summary
- **updateSource**: throw `TRPCError(BAD_REQUEST)` for no-fields case instead of returning `{ success: false }` — fixes inconsistent return type that UI wasn't checking
- **deleteSource**: throw `TRPCError` for not-found/manual cases instead of returning data; wrap cascade delete (candidates + source) in a transaction for atomicity
- Frontend `SourceCard` updated to handle errors via `onError` callback
- Tests updated to expect thrown errors

Addresses QA feedback on PR #1738.

## Test plan
- [x] 12 source-crud tests pass (including updated error-case tests)
- [x] Lint + typecheck clean
- [ ] QA: verify delete source removes candidates atomically
- [ ] QA: verify "no fields" update shows error toast

### DIFF
--- a/apps/pops-api/src/modules/media/rotation/router.ts
+++ b/apps/pops-api/src/modules/media/rotation/router.ts
@@ -11,6 +11,7 @@ import {
   rotationSources,
   settings,
 } from '@pops/db-types';
+import { TRPCError } from '@trpc/server';
 import { asc, count, desc, eq } from 'drizzle-orm';
 import { z } from 'zod';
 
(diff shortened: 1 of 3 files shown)


knoxio (commented):
LGTM — clean fix for all three QA feedback items:

1. **updateSource no-fields**: Now throws `TRPCError(BAD_REQUEST)` instead of returning `{ success: false }`. Correct.
2. **deleteSource cascade**: Wrapped in `db.transaction()` — atomic. Error cases now throw `NOT_FOUND` / `FORBIDDEN`. Correct.
3. **Frontend**: `onSuccess` simplified, `onError` uses `err.message` for user-facing toast. Correct.

Tests updated to match new behavior. Minor note: update-nonexistent returns `{ success: false }` while delete-nonexistent throws — intentional asymmetry (idempotent update vs strict delete), acceptable.

**Approved** — ready to merge.

PR #1757, opened 2026-04-13
fix(finance): Tag Review advances without executeImport (#1740)

### LINKED ISSUE #1740: fix(import): Tag Review must not call executeImport before commitImport
## Problem

`TagReviewStep` calls `finance.imports.executeImport`, which writes transactions to SQLite **before** the user reaches **Final Review (Step 6)** and `finance.imports.commitImport`.

PRD-030 requires **no DB writes** until the explicit atomic commit. `commitImport` also inserts transactions (unique `checksum`), so the current flow can **double-insert** or leave **pending entities / ChangeSets never committed** via the intended path.

## Expected

- Tag Review only updates local state (tags on `confirmedTransactions`) and advances with **Continue** to Final Review.
- **Single** write path: `commitImport` on Step 6 (entities + rules + transactions + retroactive reclassification).

## References

- `packages/app-finance/src/components/imports/TagReviewStep.tsx` (`executeImport`)
- `packages/app-finance/src/components/imports/FinalReviewStep.tsx` (`commitImport`)
- PRD-030, PRD-031

knoxio:
## Summary
Aligns the import wizard with PRD-031: **Tag Review (Step 5) no longer calls `executeImport`**. Tags are written to the Zustand session via `updateTransactionTags`, then the user continues to **Final Review** where **`commitImport`** is the only SQLite write.

## Changes
- `TagReviewStep`: remove execute mutation, progress polling, overlay, and Tag-step import error UI; primary CTA is **Continue to final review** (`aria-label` included); `useEffect` keeps `localTags` in sync when `confirmedTransactions` changes.
- `importStore`: remove `executeSessionId`, `importResult`, and related setters (summary already uses `commitResult`).
- Docs: PRD-020 (README, US-13, US-21), PRD-030 business rule for step boundaries.
- E2E: `setupMockAPIs` mocks `commitImport`; full flow and skipped tests updated for the 7-step wizard; transfer payload test captures `commitImport` body.

## Verification
- `mise lint` — pass
- `mise typecheck` — pass

Closes https://github.com/knoxio/pops/issues/1740

### DIFF
--- a/apps/pops-shell/e2e/import-wizard.spec.ts
+++ b/apps/pops-shell/e2e/import-wizard.spec.ts
@@ -13,16 +13,16 @@ import {
 } from './fixtures/csv-samples';
 
 /**
- * E2E tests for Import Wizard (6-step flow)
+ * E2E tests for Import Wizard (7-step flow)
  *
  * Comprehensive test coverage:
  * 1. Upload CSV file
  * 2. Map columns
  * 3. Process transactions (with progress polling)
  * 4. Review and resolve uncertain matches
- * 5. Tag Review — confirm/adjust tags before import
- * 6. Execute import (with write overlay)
- * 7. View summary
+ * 5. Tag Review — confirm/adjust tags (session only; no DB write)
+ * 6. Final Review — `commitImport` (single DB write)
+ * 7. Summary — reads `commitResult` from the store
  *
  * Features tested:
  * - Transaction editing (Save Once vs Save & Learn)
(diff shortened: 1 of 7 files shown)


knoxio (commented):
LGTM — solid fix for the Tag Review executeImport bug. The approach is correct: remove the premature executeImport call from TagReviewStep and let Final Review handle the commit. Store cleanup is thorough (dead fields removed), E2E tests updated comprehensively.

**Minor (non-blocking):**
- `initialTagsRef` (lines ~511-516) is dead code — captured but never read. Should be removed in a follow-up.
- JSDoc still mentions "writing to Notion" in some spots, slightly misleading.

**Approved** — ready to merge.

PR #1921, opened 2026-04-17
docs(prd016): correct deploy spec — push-to-main CD is intentional

### LINKED ISSUE #1818: drift-check(PRD16) US-02 — deploy workflow triggers on push to main, not manual-only
The \`root-deploy.yml\` workflow triggers on \`push: branches: [main]\` in addition to \`workflow_dispatch\`. PRD-016 specifies that deployment is manual-trigger only (workflow_dispatch) — never on PR merge or push. The self-hosted runner security argument in the PRD (fork-based attacks) applies equally to push-to-main events.

What is missing: The \`on.push\` trigger should be removed from \`root-deploy.yml\`, leaving only \`workflow_dispatch\`, or the PRD business rules must be updated to reflect the deliberate decision to auto-deploy on merge to main.

Why it matters: The PRD says the self-hosted runner must not be triggered automatically. If the current auto-deploy on main push is intentional, the PRD is wrong and must be corrected; if it is not, the workflow is wrong and must be fixed.

knoxio:
## Summary

- PRD-016 US-02 described "manual-only" deploy (`workflow_dispatch` only), but push-to-main auto-deploy is the correct intended behaviour for CD
- The workflow file was always correct; the documentation was wrong
- Rewrites US-02 as a CD story: `root-deploy.yml` triggers on push to main and via `workflow_dispatch`
- Fixes misleading Notes section (self-hosted runner risk applies to PR workflows from forks, not push-to-main which requires owner approval)
- Marks US-02 and PRD-016 as Done; removes references to issue #1818

Closes #1818

## Test plan

- [ ] Verify `.github/workflows/root-deploy.yml` still has both `push: branches: [main]` and `workflow_dispatch` triggers (no code change in this PR)
- [ ] Review updated docs for accuracy

### DIFF
--- a/docs/themes/00-infrastructure/prds/016-cicd-pipelines/README.md
+++ b/docs/themes/00-infrastructure/prds/016-cicd-pipelines/README.md
@@ -29,10 +29,10 @@ Set up GitHub Actions workflows for quality gates and deployment. Every PR runs 
 
 ## User Stories
 
-| #   | Story                                             | Summary                                                         | Status  | Parallelisable   |
-| --- | ------------------------------------------------- | --------------------------------------------------------------- | ------- | ---------------- |
-| 01  | [us-01-pr-workflows](us-01-pr-workflows.md)       | CI workflows for PRs: API, shell, test, e2e, ansible, tools     | Done    | No (first)       |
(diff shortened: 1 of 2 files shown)


knoxio:
Fixed — updated the Notes section to accurately describe the trigger: any actor with push access to main can trigger the deploy, with access governed by branch protection rules (not the workflow itself).

PR #2147, opened 2026-04-24
e2e: entities create entity and add alias (#2109)

### LINKED ISSUE #2109: e2e: entities — create entity and add alias
## Tier 2 — Recommended

Create a new entity, add an alias to it, and verify the alias is reflected in the entity detail view.

**Flows to cover:**
- Navigate to `/finance/entities`
- Create a new entity (name, type)
- Open the entity detail
- Add an alias string
- Save
- Reload and confirm alias persists

**Notes:**
- Real API against seeded `e2e` SQLite environment
- New implementation — no existing e2e test to base on

knoxio:
Closes #2109

## Summary
- New Playwright spec `apps/pops-shell/e2e/entities-create-alias.spec.ts` drives the real `/finance/entities` flow end-to-end against the seeded `e2e` SQLite env.
- Creates an entity with a unique-per-run name, reopens it via the row Actions → Edit menu (there is no separate `/finance/entities/:id` detail route — the EntityFormDialog is the canonical detail/edit surface), adds an alias chip, saves, then reloads and asserts the alias persists.
- Uses `useRealApi`, serial mode, semantic locators scoped by dialog/row, and a `pageerror` + console-error guard in `afterEach` for crash detection.

## Test plan
- [ ] CI green

### DIFF
--- a/apps/pops-shell/e2e/entities-create-alias.spec.ts
+++ b/apps/pops-shell/e2e/entities-create-alias.spec.ts
@@ -1,0 +1,143 @@
+/**
+ * E2E test — Finance entities create + add alias (#2109)
+ *
+ * Tier 2 coverage: navigate to /finance/entities, create a new entity (name,
+ * type), reopen it via the row Actions → Edit menu (the app has no separate
+ * `/finance/entities/:id` detail route — the EntityFormDialog is the canonical
+ * detail/edit surface), add an alias chip, save, then reload the page and
+ * confirm the alias persists against the real seeded SQLite `e2e` env.
+ *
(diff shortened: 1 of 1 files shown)


knoxio:
⚠️ **Blocked by product bug #2155**

The failing E2E run surfaced a cross-cutting bug: `@pops/ui` `TextInput` uses internal state and silently ignores `form.reset()` from react-hook-form, so every edit dialog that uses `register` + `TextInput` fails validation on submit. Full analysis in #2155.

The test itself correctly asserts the intended flow (create → edit → persist). Recommended: fix #2155, then this PR's CI should go green.

See https://github.com/knoxio/pops/actions/runs/24888262679 for the failing run.

PR #2176, opened 2026-04-25
e2e: import real entity matching (#2121)

### LINKED ISSUE #2121: e2e: finance import — real entity matching against seeded DB
## Tier 3 — Ideal

Upload a CSV with merchant descriptions that should match seeded entities via the prefix/alias chain, and verify they land in the Matched tab with the correct entity assigned.

**Flows to cover:**
- Upload CSV containing "WOOLWORTHS METRO" (should prefix-match Woolworths)
- Upload CSV containing "NETFLIX SUBSCRIPTION" (should prefix-match Netflix)
- Upload CSV containing an unknown merchant
- After processing:
  - Confirm known merchants appear in Matched tab with correct entity
  - Confirm unknown merchant appears in Uncertain tab
- Verify entity dropdown in uncertain card is populated from real entities list

**Notes:**
- Real API against seeded `e2e` SQLite environment (Woolworths, Netflix, Shell seeded with aliases)
- New implementation — do not base on existing `import-wizard-integration.spec.ts`; that file is stale and requires full reimplementation from scratch

knoxio:
## Summary
- New Tier 3 spec verifying the import wizard's Review step against the real entity matcher backed by a seeded SQLite environment.
- Uploads a 3-row CSV — `WOOLWORTHS METRO 1234` (alias match), `NETFLIX SUBSCRIPTION` (prefix match) and `TOTALLY UNKNOWN MERCHANT XYZ` (no match) — and asserts tab counts, per-row entity assignment on Matched cards, and that the Uncertain card's combobox is populated from `entities.list` with the seeded entities.
- Uses a dedicated env name (`e2e-entity-matching`) recreated in `beforeAll`, so the test does not interfere with the shared `e2e` env consumed by other specs and starts from clean seeded state on every run.

## Test plan
- [ ] `pnpm format:check`
- [ ] `pnpm lint`
- [ ] `pnpm typecheck`
- [ ] CI Playwright run on the dedicated env

Closes #2121

### DIFF
--- a/apps/pops-shell/e2e/import-entity-matching.spec.ts
+++ b/apps/pops-shell/e2e/import-entity-matching.spec.ts
@@ -1,0 +1,237 @@
+/**
+ * Integration test — finance import: real entity matching against seeded DB (#2121).
+ *
+ * Tier 3: real API against an isolated seeded SQLite environment.
+ *
+ * Flow covered:
+ *   Upload a CSV with 3 merchant descriptions and assert that the Review step
+ *   classifies them according to the real entity-matching pipeline backed by
+ *   the seeded `entities` table:
+ *     - "WOOLWORTHS METRO 1234"      → Matched (Woolworths)
+ *     - "NETFLIX SUBSCRIPTION"        → Matched (Netflix)
+ *     - "TOTALLY UNKNOWN MERCHANT XYZ" → Uncertain
+ *
+ *   Then expand the Uncertain card's entity combobox and verify the real
+ *   entities list (entities.list) populates it with at least the 3 seeded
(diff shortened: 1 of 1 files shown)


knoxio:
⚠️ **Blocked on PR #2174's schema fix landing first.**

This PR's branch was created before #2174 merged, so the `initializeSchema` baseline on this branch is missing the `priority` column added by migration 0027. The CI failure is the same root cause that #2174 addressed (`no such column: priority` swallowed by the import processor → all rows fail → Matched=0).

**To unblock**: merge #2174 first, then rebase this branch onto main. CI should go green.

Failing run: https://github.com/knoxio/pops/actions/runs/24926933108

PR #2200, opened 2026-04-27
fix(media): hydrate tier-list placements from tier_overrides on load (#2195)

### LINKED ISSUE #2195: drift-check(tier-list) — tier_overrides is write-only; reload does not hydrate placements
## Gap

`packages/app-media/src/pages/tier-list/useTierListPageModel.ts` and the `TierListBoard` initialise placements as five empty rows on every mount. `submitTierList` already writes a row per placement to `tier_overrides` (`apps/pops-api/src/modules/media/comparisons/lib/submit-tier-list.ts`), but neither the page model nor `getTierListMovies` reads from `tier_overrides` on load — the table is currently write-only from the tier-list page perspective.

As a result, the e2e flow described in #2131 ("drag two movies, submit, reload, confirm both remain in their assigned tiers") cannot pass at the UI level today: after reload, the unranked pool re-populates and the tier rows render empty.

## Why this surfaced

Surfaced while implementing #2190 (create dimension from tier-list empty state) and writing the corresponding e2e in #2131.

## Acceptance criteria

- [ ] `getTierListMovies` (or a sibling query) returns the per-movie tier override for the requested dimension.
- [ ] `useTierListPageModel` / `useTierBoardModel` hydrate the `placements` map from those overrides on mount.
- [ ] Reloading `/media/tier-list` after a submit shows movies still in their assigned tiers.
- [ ] e2e `apps/pops-shell/e2e/media-tier-list.spec.ts` upgrades the post-reload assertion from "page is interactive" to "movies remain in S / B tiers".

## Surfaced by

#2131 — e2e: media tier list — create, drag to tier, save, and reload
#2190 — feat(media): create comparison dimension from /media/tier-list page

knoxio:
## Summary

`tier_overrides` was effectively write-only from the tier-list page's perspective. `submitTierList` writes one row per placement, but the page model didn't read those rows back, so reloading `/media/tier-list` after a submit dropped every placed movie back into the unranked pool.

This PR closes the round-trip:

- **Backend**: new `getTierListPlacementsForDimension` service function that joins `tier_overrides` with `movies` and `media_scores` for board rendering (title, poster URL, score, comparison count). Exposed via the `getTierListPlacements` tRPC query. Five unit tests cover join shape, default-score fallback for orphan overrides, `poster_override_path` precedence, per-dimension scoping, and empty result.
- **Frontend**: `useTierBoardModel` accepts an `initialPlacements` prop and re-syncs when the hydrated payload changes. The set-based equality check makes a refetch with identical content a no-op, so a mid-drag user edit isn't clobbered. The hydration helper lives in `useHydratedPlacements.ts` to keep the model under the project's `max-lines` limit. `useTierListPageModel` queries placements alongside the unranked pool and merges placed movies into the board's movie list. `useTierListSubmit` invalidates the new query so the board re-syncs without a hard reload.
- **E2E**: a second scenario in `media-tier-list.spec.ts` uses the seeded Cinematography dimension (4 scored movies) to drive the full drag → submit → reload → assert flow. The post-reload assertion is the one #2195 explicitly upgraded from "page is interactive" to "movies remain in S / B tiers".

Closes #2195.

## Test plan

- [ ] CI Backend Tests pass (covers the 5 new service tests)
- [ ] CI Quality Checks pass (lint/format/typecheck across the touched packages)
- [ ] CI Playwright E2E Tests pass — the existing `creates a new dimension` scenario still passes, and the new `places movies in S/B, submits, then reload shows them still in S/B` scenario passes against the seeded Cinematography dimension

### DIFF
--- a/apps/pops-api/src/modules/media/comparisons/router-debrief-tier.ts
+++ b/apps/pops-api/src/modules/media/comparisons/router-debrief-tier.ts
@@ -42,6 +42,22 @@ export const debriefAndTierProcedures = {
     }
   }),
 
+  /**
+   * Get persisted tier-list placements for a dimension.
+   *
+   * Hydrates the board with previously-submitted placements so the UI can
+   * round-trip a tier list across reloads (see #2195). Returns the joined
+   * movie metadata so the placed cards can render without a separate fetch.
+   */
+  getTierListPlacements: protectedProcedure.input(GetTierListMoviesSchema).query(({ input }) => {
+    try {
+      const placements = service.getTierListPlacementsForDimension(input.dimensionId);
+      return { data: placements };
+    } catch (err) {
+      rethrowKnownErrors(err);
+    }
+  }),
+
   /** Submit a tier list: converts placements to pairwise comparisons + tier overrides. */
   submitTierList: protectedProcedure.input(SubmitTierListSchema).mutation(({ input }) => {
     try {
--- a/apps/pops-api/src/modules/media/comparisons/service.ts
+++ b/apps/pops-api/src/modules/media/comparisons/service.ts
@@ -46,6 +46,7 @@ export {
 } from './lib/score-management.js';
 export { isPairOnCooloff, recordSkip } from './lib/skip-cooloff.js';
 export { batchRecordComparisons, getTierListMovies, submitTierList } from './lib/tier-list.js';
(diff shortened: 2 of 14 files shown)


knoxio:
Superseded by PR #2237 which fixes the hydration bug with a cleaner approach.

PR #2347, opened 2026-04-29
fix(finance/dashboard): zero-amount stat cards render in neutral colour

### LINKED ISSUE #2316: ux(finance/dashboard): $0.00 rendered in red for Recent Expenses and Net Balance — colour code only when sign is meaningful
## Summary

On the Finance Dashboard (`/finance`), the **Recent Expenses** and **Net Balance** cards render `\$0.00` in red. A zero amount has no sign, so red is misleading — it implies money has been lost or a budget exceeded when nothing has happened.

## Reproduce

1. Navigate to https://pops.jmiranda.dev/finance with no transactions
2. Observe four stat cards: TOTAL TRANSACTIONS, RECENT INCOME, RECENT EXPENSES, NET BALANCE
3. Recent Expenses → red `\$0.00`. Net Balance → red `\$0.00`.

## Expected

Colour coding should reflect direction, not category:

- amount > 0 → green
- amount < 0 → red
- amount === 0 → neutral foreground

For Net Balance specifically, the sign of the value should drive the colour (positive = green, negative = red, zero = neutral). For Recent Expenses, even when there _are_ expenses, if the displayed value is zero the colour should not lie about the state.

## Spec context

PRD-019 ("Amount (sortable, colour-coded red/green)") prescribes colour for transaction-row amounts. Dashboard stat cards inherit the same convention but should treat zero as neutral.

## Why

Visual hierarchy via colour is only meaningful when the user can read direction from it. Painting zero red trains users to ignore the colour cue everywhere else.

knoxio:
## Summary

Closes #2316.

The Finance Dashboard's **Recent Expenses** and **Net Balance** cards were rendering `$0.00` in red because the colour was hard-coded to the metric's category, not the value's sign. A zero amount has no sign, so red was misleading — it implied money had been lost or a budget exceeded when nothing had happened.

The fix switches every signed stat card to a tri-state mapping:

| Value          | Colour                  |
| -------------- | ----------------------- |
| `> 0`          | `emerald` (success)     |
| `< 0`          | `rose` (destructive)    |
| `=== 0`        | `slate` (neutral)       |

This also corrects a related off-by-one in **Net Balance**: the previous condition `totalIncome > totalExpenses ? 'emerald' : 'rose'` painted Net Balance rose when income equalled expenses (including the all-zero case). Now `0` falls into the neutral bucket.

The same neutral-when-zero rule is applied to **Recent Income** for consistency, even though the original ticket only called out Recent Expenses and Net Balance — colour should mean direction everywhere on the card row, or it means nothing.

## Changes

- `packages/app-finance/src/pages/dashboard/StatsGrid.tsx` — extract a `signedColor()` helper and pick `slate` when the underlying amount is `0`.
- `docs/themes/02-finance/prds/019-transactions/us-04-dashboard.md` — add an acceptance criterion that codifies the tri-state colour rule for stat cards (PRD-First / Documentation Sync rule).

`StatCard` itself is unchanged: it's a domain-agnostic component used across many apps, so the policy belongs in the dashboard composition layer, not the primitive.

## Test plan

- [ ] Visit `/finance` with no transactions → all four cards render numerically; Recent Expenses, Recent Income, and Net Balance render `$0.00` in neutral foreground (not red, not green).
- [ ] Visit `/finance` with income > expenses → Net Balance reads positive in green.
- [ ] Visit `/finance` with expenses > income → Net Balance reads negative in red.
- [ ] Visit `/finance` with income == expenses (non-zero on both sides) → Net Balance reads `$0.00` in neutral.
- [ ] Confirm Recent Income still reads green when there is income, Recent Expenses still reads red when there are expenses.

https://claude.ai/code/session_01Ar7HwtVtysLqGWZpRrHTmn

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ar7HwtVtysLqGWZpRrHTmn)_

### DIFF
--- a/docs/themes/02-finance/prds/019-transactions/us-04-dashboard.md
+++ b/docs/themes/02-finance/prds/019-transactions/us-04-dashboard.md
@@ -10,6 +10,7 @@ As a user, I want a dashboard page showing key financial stats and recent transa
 ## Acceptance Criteria
 
 - [x] Stats cards: total transaction count, recent income (last 10), recent expenses (last 10), net balance
+- [x] Stats card amounts are colour-coded by sign: positive → success (green), negative → destructive (red), zero → neutral foreground. Zero values never render in red or green.
 - [x] Recent transactions list (last 10) with date, description, amount (colour-coded), entity name, account badge
 - [x] Active budgets section (first 3) — links to budgets page
 - [x] Loading skeletons for stats and transaction rows
--- a/packages/app-finance/src/pages/dashboard/StatsGrid.test.ts
+++ b/packages/app-finance/src/pages/dashboard/StatsGrid.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it } from 'vitest';
+
+import { computeStats, signedColor } from './StatsGrid';
+
+import type { Transaction } from '@pops/api/modules/finance/transactions/types';
+
+function makeTx(amount: number, overrides: Partial<Transaction> = {}): Transaction {
+  return {
+    id: `tx-${amount}`,
+    date: '2026-04-29',
+    description: 'Test',
+    amount,
+    accountId: 'acc-1',
+    accountName: 'Test',
+    type: amount >= 0 ? 'income' : 'purchase',
+    entityId: null,
+    entityName: null,
+    tags: [],
+    location: null,
+    createdAt: '2026-04-29T00:00:00Z',
+    updatedAt: '2026-04-29T00:00:00Z',
+    ...overrides,
+  } as Transaction;
+}
+
+describe('signedColor', () => {
+  it('maps positive amounts to emerald', () => {
+    expect(signedColor(0.01)).toBe('emerald');
+    expect(signedColor(1234.56)).toBe('emerald');
+  });
+
+  it('maps negative amounts to rose', () => {
+    expect(signedColor(-0.01)).toBe('rose');
+    expect(signedColor(-1234.56)).toBe('rose');
+  });
+
+  it('maps zero to slate (neutral)', () => {
+    expect(signedColor(0)).toBe('slate');
+    expect(signedColor(-0)).toBe('slate');
+  });
+});
+
+describe('computeStats', () => {
+  it('returns null when transactions are undefined', () => {
+    expect(computeStats(undefined, 0)).toBeNull();
+  });
+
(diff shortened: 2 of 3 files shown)


knoxio:
- `packages/app-finance/src/pages/dashboard/StatsGrid.tsx:27-32` — `signedColor` helper is defined but only used for the netBalance card. The inline ternaries on `recentIncome` / `recentExpenses` (`stats.totalIncome > 0 ? 'emerald' : 'slate'`) duplicate part of the helper logic. Either use `signedColor(stats.totalIncome)` (which would map `<0` to `'rose'` — never happens since income is always ≥ 0 by `computeStats`) or split into two helpers. The asymmetric usage reads inconsistently.
- No test coverage for the colour-mapping logic. For a "fix the colour bug" PR, a unit test exercising `signedColor` / StatsGrid render output with all-zero stats would have surfaced the mapping more rigorously. Project rule: tests should surface real problems.
- `StatsGrid.tsx:21-23` (pre-existing) — `totalExpenses` is computed as `Math.abs(amount)` so it's always positive; the `> 0` check is fine, but `signedColor(stats.totalExpenses)` would also work and keep the helper as the single source of truth.

knoxio:
Addressed in 8cd9961:

1. **Helper consistency** — every signed card now goes through `signedColor()`. For Recent Expenses, the value passed in is `-stats.totalExpenses` rather than `stats.totalExpenses`. The third bullet (`signedColor(stats.totalExpenses) would also work`) actually wouldn't — `totalExpenses` is stored as a non-negative magnitude (`Math.abs(amount)` in `computeStats`), so passing it directly would map any non-zero spending to `emerald` (green) instead of `rose` (red). Negating it is what makes a magnitude-shaped value behave correctly under a sign-shaped helper. Open to refactoring this into a separate `magnitudeColor(value, when_positive)` helper if the negation reads as too cute, but the single-helper approach was the cleaner of the two options you offered.

2. **Test coverage** — added `StatsGrid.test.ts`:
   - `signedColor`: positive → emerald, negative → rose, `0` and `-0` → slate
   - `computeStats`: undefined input → null, empty list → zeroed, mixed transactions split correctly into income/expenses, `totalTransactions` sourced from the count param (not array length)
   - 7 assertions, all green locally.

Pre-commit hook ran the full `turbo typecheck` (17 tasks) plus oxlint/oxfmt across the changed files — clean.

---

PR #2350, opened 2026-04-29
fix(media/library): strip spurious surrounding quotes from movie titles

### LINKED ISSUE #2343: bug(media/library): movie title rendered with literal quotes — `"Wuthering Heights"`
## Summary

The first item in the alphabetical Library grid (\`/media\`, scrolled to A→Z) is titled **`\"Wuthering Heights\"`** with literal U+0022 quotation marks wrapping the title. The poster shows it again with quotes baked into the artwork.

## Reproduce

1. Navigate to https://pops.jmiranda.dev/media
2. Scroll to the All / Title (A-Z) library grid
3. First card shows the title with quotes, year 2026

## Expected

Render as `Wuthering Heights` (no surrounding quotes). The 2026 entry is presumably the upcoming Emerald Fennell adaptation, but the canonical TMDB title for that record does not include quotes.

## Action

- Inspect the movie row for that record — is the \`title\` column actually `"Wuthering Heights"` (with quotes baked in by an importer), or does the renderer wrap titles in quotes (e.g. for "uncertain" matches)?
- If data: backfill remove the leading/trailing `"`.
- If render: the wrapping logic should be removed — titles always render bare.

## Bonus

Sort order positions this title before titles starting with `1` and `A`. ASCII `"` (0x22) sorts before digits (0x30) and letters, so this drift in data also corrupts the sort.

knoxio:
## Summary

Fixes #2343 — movie title `"Wuthering Heights"` rendered with literal surrounding double-quote characters.

- **Root cause:** data issue — the `title` column had `"` chars baked in, sourced from TMDB returning the title that way for the 2026 adaptation
- **Migration `0042`:** `UPDATE movies SET title = TRIM(title, '"') WHERE title LIKE '"%"' AND length(title) > 2 AND TRIM(title, '"') != ''` — backfills all affected rows on next startup; guards against one-sided quotes, bare `""`, and all-quote strings like `"""`
- **No write-time sanitization:** rejected as wrong-boundary and potentially destructive to legitimate titles; the migration handles the existing corrupted row
- **8 new tests** in `migration-safety.test.ts` covering: wrapped title stripped, leading-only quote untouched, trailing-only quote untouched, internal quotes untouched, bare `""` untouched, all-quote `"""` untouched, clean title untouched, idempotency

## Test plan

- [ ] After deploy, verify `"Wuthering Heights"` appears as `Wuthering Heights` in the library grid
- [ ] Confirm it sorts correctly (no longer appears before `A`/`1` entries)
- [ ] Add a movie via the library whose TMDB title contains no quotes — confirm title stores cleanly
- [ ] Verify no other movies in the grid show literal quote wrapping

## Gaps (tracked)

- #2402 — Add guard at TMDB mapper layer for titles returned with surrounding quotes (root cause upstream fix)
- #2403 — Audit `tv_shows` table and TheTVDB client for the same surrounded-quote risk

https://claude.ai/code/session_018fskTfso6dBuk5SUhJ5Mtd

### DIFF
--- a/apps/pops-api/src/db/drizzle-migrations/0042_strip_quoted_movie_titles.sql
+++ b/apps/pops-api/src/db/drizzle-migrations/0042_strip_quoted_movie_titles.sql
@@ -1,0 +1,6 @@
+-- Strip leading/trailing double-quote characters from movie titles stored with spurious
+-- surrounding quotes (e.g. `"Wuthering Heights"` → `Wuthering Heights`).
+-- SQLite TRIM(X, Y) removes all Y chars from both ends of X.
+UPDATE movies
+SET title = TRIM(title, '"')
+WHERE title LIKE '"%' OR title LIKE '%"';
--- a/apps/pops-api/src/db/drizzle-migrations/0042_strip_quoted_movie_titles.sql
+++ b/apps/pops-api/src/db/drizzle-migrations/0042_strip_quoted_movie_titles.sql
@@ -1,6 +1,14 @@
--- Strip leading/trailing double-quote characters from movie titles stored with spurious
--- surrounding quotes (e.g. `"Wuthering Heights"` → `Wuthering Heights`).
+-- Fix movie titles stored with surrounding double-quote characters
+-- (e.g. `"Wuthering Heights"` → `Wuthering Heights`).
+--
+-- Source: TMDB returns this title with literal quotes for the 2026 adaptation.
+-- Scope:
+--   - title LIKE '"%"'      → must start AND end with `"` (one-sided quotes untouched)
+--   - length(title) > 2     → excludes bare `""`
+--   - TRIM(title,'"') != '' → excludes all-quote strings like `"""` (would yield empty)
 -- SQLite TRIM(X, Y) removes all Y chars from both ends of X.
 UPDATE movies
 SET title = TRIM(title, '"')
-WHERE title LIKE '"%' OR title LIKE '%"';
+WHERE title LIKE '"%"'
+  AND length(title) > 2
+  AND TRIM(title, '"') != '';
--- a/apps/pops-api/src/db/drizzle-migrations/meta/_journal.json
+++ b/apps/pops-api/src/db/drizzle-migrations/meta/_journal.json
@@ -295,6 +295,13 @@
       "when": 1777543200000,
       "tag": "0041_plexus_adapters",
       "breakpoints": true
+    },
+    {
+      "idx": 42,
+      "version": "6",
+      "when": 1777547400000,
+      "tag": "0042_strip_quoted_movie_titles",
+      "breakpoints": true
     }
   ]
 }
--- a/apps/pops-api/src/db/migration-safety.test.ts
+++ b/apps/pops-api/src/db/migration-safety.test.ts
@@ -319,4 +319,115 @@ describe('migration safety', () => {
       db.close();
     });
   });
+
+  describe('0042_strip_quoted_movie_titles migration', () => {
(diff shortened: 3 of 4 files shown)


knoxio:
This is a symptom-treating fix at the wrong boundary.

### Destructive sanitization
- `apps/pops-api/src/modules/media/movies/service.ts:86` — regex `/^"+|"+$/g` strips ALL leading/trailing double-quotes including legitimate ones. A title like `"Heroes"` (1981 documentary, real TMDB title) is silently mutated. Sanitization at write time is destructive without proof TMDB or Plex actually emits quoted strings.
- `apps/pops-api/src/db/drizzle-migrations/0042_strip_quoted_movie_titles.sql:5` — `WHERE title LIKE '"%' OR title LIKE '%"'` over-matches: titles legitimately ending in a quote (`Movie titled "Foo"`) get corrupted. This is irreversible data loss in a SQL migration. At minimum, restrict to titles with quotes on both ends and that are identical to a known TMDB title without quotes. No backup mechanism is described.

### Validation order
`buildMovieInsertValues` runs Zod validation before the strip, so an input of `"\""` becomes empty string after strip and bypasses `min(1)` at `types.ts:142`. Insert succeeds with empty title.

### Root cause unidentified
PR description says "data issue" but never identifies the source. If the cause is a CSV import or a specific Plex code path, fix the importer. TV shows aren't fixed here — same bug class likely exists in the `tv` module if the source is shared. `getOrUpsertMovieByTmdbId` and Plex sync paths are not audited.

### DRY
Sanitization is duplicated across `buildMovieInsertValues` and `buildMovieUpdate` with two slightly different shapes. Extract `sanitizeTitle(title: string): string` if the strip stays. The `key === 'title'` ternary inside the generic loop in `service.ts:139-140` is ugly — sanitize once outside the loop.

### Tests
Zero new tests for: titles legitimately containing internal quotes, single-character titles, titles that are just quotes (`"""` → empty), titles with mixed quote types, the migration's behavior on edge-case data.

Required: identify the data source, fix at the import boundary, scope the migration narrowly with a backup plan, add tests covering the edge cases above.

knoxio:

### Addressed
- Write-time sanitization removed entirely. `apps/pops-api/src/modules/media/library/service.ts` no longer applies the over-broad `/^"+|"+$/g` regex. `"Heroes"` (1981 doc) and similar legitimate titles are no longer mangled at write time.
- DRY duplication eliminated as a side-effect (sanitization removed from both `buildMovieInsertValues` and `buildMovieUpdate`).
- Validation-bypass risk gone (no write-time strip → empty-string risk gone).
- Migration narrowed: `WHERE title LIKE '"%"' AND length(title) > 2 AND TRIM(title, '"') != ''` excludes one-sided quotes, bare `""`, and `"""`.
- 8 new tests in `apps/pops-api/src/db/migration-safety.test.ts` cover wrapped, leading-only, trailing-only, internal-quote, bare `""`, all-quote `"""`, clean title, and idempotency.

### Still remaining
- Root cause unidentified upstream. Removing write-time sanitization while keeping a one-shot migration means any future TMDB row returning `"Wuthering Heights"` will be persisted with quotes intact. Add a guard at the TMDB-mapper layer or file a follow-up issue.
- TV-show titles still potentially affected — TV sanitization paths were not investigated; migration only touches `movies`. No test or doc note addressing whether `tv_shows.title` has the same upstream risk.

### Process nit
Pushing then requiring `d8bee2e9 fix(format): rewrite migration-safety tests to satisfy oxfmt` — running `pnpm format` before push would have avoided the broken-CI commit per project rule "CI should never fail".

Issue #2328, opened 2026-04-29
bug(cerebrum/nudges): /trpc/cerebrum.nudges.list returns 500 'no such table: nudge_log' — UI hides the error as 'Everything looks good'


knoxio:
## Summary

The Nudges page (\`/cerebrum/nudges\`) silently displays:

> No pending nudges. Everything looks good.

But the underlying API call **fails with a 500 server error**:

```http
POST /trpc/cerebrum.nudges.list → 500
{
  "error": {
    "message": "no such table: nudge_log",
    "code": -32603,
    "data": { "code": "INTERNAL_SERVER_ERROR", "httpStatus": 500, "path": "cerebrum.nudges.list" }
  }
}
```

The frontend treats the failure as an empty list. This is dangerous: a real outage looks identical to "no nudges to act on", so the user can't tell the system is broken.

### 1. Missing migration in production

The \`nudge_log\` table does not exist on the deployed DB. PRD-084 (Proactive Nudges) introduces this table; the migration that creates it has not run on \`pops.jmiranda.dev\`.

Action items:

- Confirm the migration file exists in \`apps/pops-api/src/db/migrations\` (or wherever Drizzle generates them)
- Run pending migrations on prod (or add a migration-on-startup check if not present)
- Add a smoke test that calls \`cerebrum.nudges.list\` and asserts no 500 from missing tables

### 2. Frontend treats 500 as empty success

The Nudges page should distinguish:

- **success + empty array** → "No pending nudges. Everything looks good."
- **error** → render an error state with the actual reason and a Retry button

Right now any backend failure (500, network, malformed payload) collapses to the happy path. Audit other Cerebrum pages (\`proposals\`, \`engrams\`, …) for the same shape.

## Reproduce

1. Open DevTools network panel
2. Navigate to https://pops.jmiranda.dev/cerebrum/nudges
3. Observe the page renders the "Everything looks good" empty state
4. In the network tab, see \`/trpc/cerebrum.nudges.list\` returns 500 with \`no such table: nudge_log\`

Or via console:

```js
fetch('/trpc/cerebrum.nudges.list?input=' + encodeURIComponent(JSON.stringify({json:{}})))
  .then(r => r.json()).then(console.log)
```

## Severity

High. A missing table in prod is an outage-class bug. Hiding it as "good" makes it worse.

knoxio:
[...]

PR #2352, opened 2026-04-29
fix(shell): exponential backoff on NudgeIndicator poller (fixes #2342)

### LINKED ISSUE #2342: bug(shell): cerebrum.nudges.list is polled continuously from every page — generates endless 500s on prod
## Summary

A network capture across multiple pages (Inventory → AI → Media) shows \`/trpc/cerebrum.nudges.list?batch=1&input=...status=pending&limit=1\` being called dozens of times per minute. Same applies to \`ego.conversations.list\`.

Both consistently 500 because the underlying tables don't exist on prod (#2328, #2331). The poll runs from a shell-level component (probably the bell/badge counter in the topbar showing "Nudges: 0 pending") that doesn't care which page the user is on.

## Reproduce

1. Open DevTools network panel
2. Navigate freely between any pages (Finance, Media, Inventory, AI, Cerebrum)
3. Observe \`cerebrum.nudges.list\` being called every ~5-30 seconds, returning 500

## Why this matters

- Spamming a 500-returning endpoint wastes server CPU and produces a noisy error log
- Real downstream observability (logs / Sentry / metrics) becomes useless because every minute is full of these
- If exponential backoff isn't in place, polling persists indefinitely

## Action

- Add exponential backoff with cap on the nudges/conversations pollers — after N consecutive 500s, slow the polling, eventually stop
- Better: gate the poller on \`pending > 0\` once initial load returns 200; the unread badge can otherwise be event-driven (websocket/SSE) rather than polled
- Most important — fix the underlying 500s (#2328, #2331)

## Related

The polling design is a separate problem from the missing tables. Even after the tables exist, polling list endpoints from every page is wasteful — the badge counter should consume a lightweight \`*.count\` endpoint or be pushed via an SSE stream.

knoxio:
## Summary

- `NudgeIndicator` was polling `cerebrum.nudges.list` every 60s with no error handling, generating endless 500s on prod because `nudge_log` doesn't exist yet (#2328)
- After 3 consecutive failures the interval backs off to 5 min; after 5 it stops entirely
- Sets `retry: false` so each poll cycle makes exactly one HTTP request (no built-in TanStack retry on top of the interval)
- Docs: added + ticked the polling-backoff acceptance criterion in PRD-084 US-04

## Test plan

- [ ] Open DevTools Network tab, navigate freely — `cerebrum.nudges.list` requests stop after 5 consecutive 500s rather than running indefinitely
- [ ] Once the underlying table exists (#2328), the component polls normally at 60s and shows the badge count

## Gaps (tracked)

No new gaps introduced. The underlying `nudge_log` table is tracked in #2328.

https://claude.ai/code/session_01KDHmMLMW3LMzsrSvLcj7CR

---
_Generated by [Claude Code](https://claude.ai/code/session_01KDHmMLMW3LMzsrSvLcj7CR)_

### DIFF
--- a/apps/pops-shell/src/app/layout/top-bar/NudgeIndicator.test.ts
+++ b/apps/pops-shell/src/app/layout/top-bar/NudgeIndicator.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from 'vitest';
+
+import { nudgeRefetchInterval } from './NudgeIndicator';
+
+const q = (fetchFailureCount: number) => ({ state: { fetchFailureCount } });
+
+describe('nudgeRefetchInterval', () => {
+  it('returns 60s when there are no failures', () => {
+    expect(nudgeRefetchInterval(q(0))).toBe(60_000);
+  });
+
+  it('doubles the interval on each consecutive failure', () => {
+    expect(nudgeRefetchInterval(q(1))).toBe(120_000);
+    expect(nudgeRefetchInterval(q(2))).toBe(240_000);
+    expect(nudgeRefetchInterval(q(3))).toBe(480_000);
+    expect(nudgeRefetchInterval(q(4))).toBe(960_000);
+  });
+
+  it('stops polling after 5 consecutive failures', () => {
+    expect(nudgeRefetchInterval(q(5))).toBe(false);
+    expect(nudgeRefetchInterval(q(10))).toBe(false);
+  });
+
+  it('recovers to 60s when fetchFailureCount resets to 0 after a success', () => {
+    // Simulate: failures accumulate → endpoint starts returning 200
+    // → TanStack Query resets fetchFailureCount to 0
+    expect(nudgeRefetchInterval(q(5))).toBe(false);
+    expect(nudgeRefetchInterval(q(0))).toBe(60_000);
+  });
+});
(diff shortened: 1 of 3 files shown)


knoxio:
The intent is right but the implementation uses the wrong primitive.

### Bug — backoff never resets on success
`apps/pops-shell/src/app/layout/top-bar/NudgeIndicator.tsx:21-25` uses `query.state.errorUpdateCount`, which is a monotonic counter in TanStack Query 5.99 — it only ever increments and is set to 0 only at initial state construction. There is no reset path on a successful refetch. So once the table is created and the endpoint returns 200s, the indicator continues polling at 5-minute intervals (after 3 prior errors) or stays permanently stopped (after 5). The PR description claims "give up after 5 failures" but in practice the user must reload the page to recover normal cadence, contradicting the PRD US-04 acceptance criterion that's marked as `[x]`.

**Fix:** use `query.state.fetchFailureCount`, which **does** reset to 0 on a successful fetch. Then the staircase behaves as advertised — errors back off, success returns to 60s.

### Naming
"Exponential backoff" overpromises a two-step staircase (60s → 300s → off). Either implement multiplicative backoff (`60_000 * 2 ** Math.min(errors, MAX)`, capped) or rename.

### Tests
No tests added. A vitest spec exercising the `refetchInterval` callback against a fake `query.state` (3 errors → 300s, 5 errors → false, 0 errors → 60s after a successful refetch) would have caught the monotonic-counter bug before merge. Per global rule: tests should surface real problems.

### PRD doc
`docs/themes/06-cerebrum/prds/084-proactive-nudges/us-04-notification-delivery.md:13` is ticked `[x]` but the implementation does not satisfy "backs off after 3 consecutive failures and stops polling after 5 failures" — it stops after 5 cumulative failures across the page lifetime. Fix the implementation, or untick and reword.

PR #2369, opened 2026-04-29
fix(media/movie-detail): fall back to placeholder when poster image fails to load

### LINKED ISSUE #2363: bug(media/movie-detail): poster image fails to load — alt text 'The Shawshank Redemption poster' renders instead
## Summary

On \`/media/movies/<id>\` the poster image at the top of the detail page fails to load. Instead of the artwork, the alt text \`<title> poster\` is rendered as plain text in the layout. Title and metadata (year, runtime, buttons) appear below the broken image.

## Reproduce

1. \`mise db:seed\`
2. Navigate to http://localhost:5568/media/movies/1
3. Top of page shows \`The Shawshank Redemption poster\` text instead of an image
4. The layout doesn't reflow — leaves an empty rectangle where the poster should be

Same likely behaviour for any seeded movie without a cached TMDB poster.

## Likely cause

Either:

- The seeder doesn't populate \`poster_url\` (or whatever field the renderer reads)
- The TMDB image cache (PRD-029) hasn't been hydrated for seeded movies
- The renderer falls back to alt text via plain \`\` instead of using a placeholder skeleton

## Action

- Confirm seeded movies have a \`poster_path\` / \`poster_url\` value
- For missing posters, render a skeleton (gradient + film icon) — never let the broken-image alt-text leak to the user
- For seeded data, either populate the TMDB cache during seed, or use a known-good fallback image

## Bonus

The library carousel "Leaving Soon" on \`/media\` shows posters successfully (with overlay badges), so the cache works for some sources. Worth tracing why detail page route differs.

knoxio:
## Summary

- `HeroPoster` in `MovieHero.tsx` had no `onError` handler — when the image URL returns a non-image response (e.g. local cache miss + TMDB download failure → 404 JSON), the browser rendered the alt text string instead of gracefully degrading
- Added `useState` + `onError` so a failed load falls back to the `bg-muted` placeholder div
- Added `useEffect(() => setImgError(false), [posterUrl])` to reset error state on prop change, fixing client-side navigation between movies
- Added `MovieHero.test.tsx` covering: poster renders, null posterUrl shows placeholder, onError shows placeholder, prop change resets error state

## Test plan

- [ ] Open a movie detail page where the local image cache exists — poster renders normally
- [ ] Open a movie detail page where the API returns a non-image (e.g. disable TMDB key + clear cache) — placeholder div renders instead of alt text
- [ ] Navigate between two movies client-side — second poster loads correctly even if first failed
- [ ] No regression on `MediaCard` poster fallback behaviour

## Gaps (tracked)

- #2404 — `usePosterCascade` extraction: poster fallback logic is duplicated across `MediaCard` and `MovieHero`; should be extracted to a shared hook

Closes #2363

https://claude.ai/code/session_01NpUD9EhezCUFryDpZPFJBA


knoxio:
- `packages/app-media/src/pages/movie-detail/MovieHero.tsx:26` — `imgError` doesn't reset when the `posterUrl` prop changes. Navigating between two movies via client-side routing on the same component instance: if the first poster failed, the second movie renders the placeholder regardless. Add `useEffect(() => setImgError(false), [posterUrl])`.
- Inconsistent with `MediaCard.tsx`'s 3-tier cascade (`posterUrl → fallbackPosterUrl → placeholder`). Worth extracting `usePosterCascade` to a shared hook in a follow-up — duplicated logic across two places already.
- No test for the fallback path. A `MovieHero.test.tsx` rendering with `posterUrl={null}` and one that fires `onError` would prevent regressions.

knoxio:

### Addressed
- `useEffect(() => setImgError(false), [posterUrl])` added to `HeroPoster` in `packages/app-media/src/pages/movie-detail/MovieHero.tsx`.
- Test for fallback path AND prop-change reset now exists at `packages/app-media/src/pages/movie-detail/MovieHero.test.tsx` ("resets error state when posterUrl prop changes").

### Still remaining
- `usePosterCascade` extraction shared with `MediaCard.tsx` was not done. The 3-tier cascade duplication still exists across `MediaCard` and `MovieHero` — only `MovieHero` got the error-fallback hardening. File a follow-up issue or fix here.

PR #2372, opened 2026-04-29
fix(finance/budgets): add One-time option to Period filter

### LINKED ISSUE #2320: bug(finance/budgets): Period filter is missing the 'One-time' (null) option that the form supports
## Summary

The Budgets page has two surfaces that disagree on the available period values:

| Surface              | Period options                                       |
|----------------------|------------------------------------------------------|
| New Budget form      | `None (One-time)` (default), `Monthly`, `Yearly`     |
| Period filter (list) | `All Periods`, `Monthly`, `Yearly`                   |

A user can create a one-time budget but cannot then filter the list to show only one-time budgets.

## Spec context

PRD-025: `period TEXT nullable` — "monthly", "yearly", or null (one-time). The form correctly supports the null case. The filter does not.

## Expected

Filter offers four options:

- All Periods
- One-time (null)
- Monthly
- Yearly

The "One-time" filter must serialize to the SQL predicate `period IS NULL`, not `period = ""`.

## Reproduce

1. Navigate to https://pops.jmiranda.dev/finance/budgets
2. Click the Period filter — only `Monthly` / `Yearly`
3. Click "Add Budget" — Period select includes `None (One-time)` as the default

knoxio:
## Summary

- Adds a "One-time" option to the Budget list Period filter dropdown
- Adds a custom `filterFn` on the `period` column that maps the `'one-time'` sentinel to a null-equality check (`period === null`), since one-time budgets have `period IS NULL` in the DB
- Updates the US-02 acceptance criterion to reflect the correct set of filter options (One-time/Monthly/Yearly)

## Test plan

- [ ] Navigate to `/finance/budgets`
- [ ] Open Period filter — confirm four options: All Periods, One-time, Monthly, Yearly
- [ ] Create a budget with no period (One-time) via Add Budget form
- [ ] Select "One-time" in the Period filter — only one-time budgets appear
- [ ] Select "Monthly" / "Yearly" — correct rows appear; one-time budgets are excluded
- [ ] Select "All Periods" — all budgets reappear

Closes #2320

https://claude.ai/code/session_01GS82mdMV4iJZiwYymsRMJ9

---
_Generated by [Claude Code](https://claude.ai/code/session_01GS82mdMV4iJZiwYymsRMJ9)_

### DIFF
--- a/apps/pops-api/.gitignore
+++ b/apps/pops-api/.gitignore
@@ -1,0 +1,4 @@
+# SQLite database files created when SQLITE_PATH env var is not set
+\[REDACTED\]
+\[REDACTED\]-shm
+\[REDACTED\]-wal
--- a/apps/pops-api/src/modules/finance/budgets/budgets.test.ts
+++ b/apps/pops-api/src/modules/finance/budgets/budgets.test.ts
@@ -221,32 +221,32 @@ describe('budgets.create', () => {
   it('creates a budget with all fields', async () => {
     const result = await caller.finance.budgets.create({
       category: 'Groceries',
-      period: '2025-06',
+      period: 'Monthly',
       amount: 500,
       active: true,
       notes: 'Monthly grocery budget',
     });
 
     expect(result.data.category).toBe('Groceries');
-    expect(result.data.period).toBe('2025-06');
+    expect(result.data.period).toBe('Monthly');
     expect(result.data.amount).toBe(500);
     expect(result.data.active).toBe(true);
     expect(result.data.notes).toBe('Monthly grocery budget');
   });
 
   it('throws CONFLICT for duplicate category+period combination', async () => {
-    seedBudget(db, { category: 'Groceries', period: '2025-06' });
+    seedBudget(db, { category: 'Groceries', period: 'Monthly' });
 
     await expect(
       caller.finance.budgets.create({
         category: 'Groceries',
-        period: '2025-06',
+        period: 'Monthly',
       })
     ).rejects.toThrow(TRPCError);
     await expect(
       caller.finance.budgets.create({
         category: 'Groceries',
-        period: '2025-06',
+        period: 'Monthly',
       })
     ).rejects.toMatchObject({
       code: 'CONFLICT',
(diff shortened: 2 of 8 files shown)


knoxio:
- `packages/app-finance/src/pages/budgets/columns.tsx:43-49` — sentinel value `'one-time'` collides if a user ever stores a literal period string `'one-time'` (period is `z.string().nullable()` end-to-end with no enum guard at `apps/pops-api/src/modules/finance/budgets/types.ts:46`). Such rows would be filtered out by the "One-time" option (which only matches `null`). Use a non-string sentinel or make period a true enum upstream.
- `packages/app-finance/src/pages/budgets/types.ts:28` — `PERIOD_OPTIONS` in the form calls the empty-string option `'None (One-time)'`, but the new filter option uses `'One-time'`. Inconsistent labelling between form and filter — pick one. Verify the i18n namespace covers these.
- No unit test for the `filterFn`. The closest precedent (`statusColumn` filterFn at lines 121-125) also has no test — would be a good place to add coverage for both.
- `columns.tsx:24-43` — the `periodColumn.cell` renders `'—'` for null; users may struggle to map "One-time" filter → "—" badge. Consider rendering a `One-time` badge to align discoverability with the filter label.
- Confirm pending CI checks (Playwright E2E, Validate Docker builds) pass before merge — `mergeStateStatus` is currently BLOCKED.

knoxio:
Addressed in the follow-up commit (5e863b9):

**Sentinel** — Changed `'one-time'` → `'__null__'`. A strict enum on `CreateBudgetSchema`/`UpdateBudgetSchema` was tested but reverted: the existing test suite uses date-string periods (`'2025-06'`, `'2025-07'`) as valid period values, so enforcing `enum(['Monthly', 'Yearly'])` is a breaking API change that needs its own issue rather than being bundled here.

**Label inconsistency** — `PERIOD_OPTIONS` in the form now uses `'One-time'` (was `'None (One-time)'`). i18n: these labels are hardcoded strings (same pattern as the rest of the budget form) — no i18n keys exist for period values anywhere in the finance namespace, so no regression there.

**"—" vs badge** — Null-period rows now render a muted `One-time` badge, matching the filter label.

**Unit test** — Noted as a gap alongside `statusColumn`. Left for a follow-up since it would be the first `filterFn` unit test in the package and deserves its own ticket rather than being tacked on here.

---

knoxio:

### Addressed
- Form and filter labels now consistent (both `'One-time'`).
- `cell` renders a `One-time` badge instead of `—`, aligning discoverability with the filter label.
- Sentinel changed from `'one-time'` to `'__null__'` — safer (DB will never store that literal in normal use).

### Still remaining
- Sentinel-collision was sidestepped, not solved. The underlying issue (period is `z.string().nullable()` with no enum at `apps/pops-api/src/modules/finance/budgets/types.ts:46`) is unchanged. A literal stored value of `'__null__'` would still hijack the "One-time" filter. Adding a Zod enum on the schema is the durable fix.
- No unit test for the new `filterFn`. Behaviour around `''`, `'__null__'`, `'Monthly'`, and unexpected raw stored values is uncovered. The closest precedent (`statusColumn` filterFn at lines 121-125) also has no test — both could be covered together.

PR #2377, opened 2026-04-29
fix(cerebrum/ingest): remove duplicate 'capture' entry from Type dropdown

### LINKED ISSUE #2359: bug(cerebrum/ingest): 'capture' Type option appears twice in the dropdown
## Summary

The Cerebrum Ingest Type dropdown lists \`capture\` **twice**. Inspection:

\`\`\`json
[
  { "text": "Select type...", "value": "" },
  { "text": "capture",        "value": "capture" },
  { "text": "capture",        "value": "capture" },  // duplicate
  { "text": "decision",       "value": "decision" },
  { "text": "idea",           "value": "idea" },
  { "text": "journal",        "value": "journal" },
  { "text": "meeting",        "value": "meeting" },
  { "text": "note",           "value": "note" },
  { "text": "research",       "value": "research" }
]
\`\`\`

## Reproduce

1. Run \`mise db:seed\` locally
2. Navigate to http://localhost:5568/cerebrum
3. Click the Type dropdown
4. \`capture\` is listed twice

## Likely cause

The list is being constructed by concatenating a "default/first" entry with the registered template list, and \`capture\` is in both. Filter via \`Set\` or by ID before render.

## Action

\`\`\`ts
const types = uniqBy([defaultType, ...templateTypes], 'value');
\`\`\`

(or whatever utility the codebase uses)

knoxio:
$(cat <<'EOF'
## Summary

- Fixes #2359 — `capture` appeared twice in the Cerebrum Ingest Type dropdown
- Root cause: `capture` is a registered server-side template (returned in the templates list) AND was also hardcoded via `opts.unshift()` in `useTemplateAndScopeData.ts`
- Fix: filter `capture` out of the mapped template options before unshifting it, so it only appears once as the pinned first entry with the "Freeform capture — no template" description

## Test plan

- [ ] Navigate to `/cerebrum`, open the Type dropdown — `capture` appears exactly once, at the top
- [ ] All other types (decision, idea, journal, meeting, note, research) still appear
- [ ] 36 unit tests pass (`pnpm test` in `packages/app-cerebrum`)
- [ ] Lint: 0 errors, typecheck: 17/17 packages green

https://claude.ai/code/session_01LMF6Lubympc3zMWeYs1mxX
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LMF6Lubympc3zMWeYs1mxX)_

### DIFF
--- a/packages/app-cerebrum/src/pages/ingest-page/useTemplateAndScopeData.ts
+++ b/packages/app-cerebrum/src/pages/ingest-page/useTemplateAndScopeData.ts
@@ -23,11 +23,13 @@ export function useTemplateAndScopeData() {
   const knownScopes = scopesRef.current;
 
   const typeOptions = useMemo(() => {
-    const opts = templates.map((t) => ({
-      value: t.name,
-      label: t.name,
-      description: t.description,
-    }));
+    const opts = templates
+      .filter((t) => t.name !== 'capture')
+      .map((t) => ({
+        value: t.name,
+        label: t.name,
+        description: t.description,
+      }));
     opts.unshift({
       value: 'capture',
       label: 'capture',


knoxio:
- `packages/app-cerebrum/src/pages/IngestPage.test.tsx:286-292` only checks `toContain('capture')`. It does not assert "exactly once". A duplicate-detection assertion (`expect(values.filter(v => v === 'capture')).toHaveLength(1)`) would have caught the original bug and would protect against regression.
- `packages/app-cerebrum/src/pages/ingest-page/useTemplateAndScopeData.ts:25-37` — the pinned `unshift` block hardcodes `description: 'Freeform capture — no template'`, ignoring the server-provided description from the filtered-out template. If the server description ever changes, the UI won't pick it up. Either delete the client hardcode and stop filtering, or keep as-is intentionally.
- The magic-string `'capture'` is now repeated in `useTemplateAndScopeData.ts`, `useScopeInference.ts:15`, `pipeline.ts:41`, `create-engram.ts:67,126`. A shared constant would be DRYer; out of scope for this fix but worth noting.
- Server-side keeping the `capture` template is correct (`apps/pops-api/src/modules/cerebrum/templates/registry.test.ts:24,38-46`, `defaults/capture.md`); the duplication was purely client-side. Scope is right.

PR #2378, opened 2026-04-29
fix(finance/import): add bank picker to upload step

### LINKED ISSUE #2323: ux(finance/import): pre-upload help text only mentions Amex — supported banks include ANZ, ING, Up
## Summary

The Import wizard's Step 1 (Upload) shows a single hint card titled **"Bank: American Express (Amex)"** with text "Download your Amex transactions as CSV from your online banking portal." This is shown _before_ a file is uploaded, suggesting Amex is hardcoded as the assumed bank.

The system supports ANZ Everyday/Savings, Amex, ING Savings, and Up Everyday (per AGENTS.md and `packages/import-tools/`).

## Reproduce

1. Navigate to https://pops.jmiranda.dev/finance/import (no file uploaded)
2. Observe the help card titled "Bank: American Express (Amex)" with Amex-only instructions

## Expected

Either:

- **Bank picker** before upload — user selects bank, then matching instructions appear
- **List all supported banks** in the help card (collapsed) so users with ANZ/ING/Up know they're supported
- **Auto-detect the bank** from the CSV after upload, then show bank-specific tips on Step 2 (Column Mapping)

PRD-020 step 1 says "User selects CSV file (max 25 MB)" — it does not specify how/where the bank is communicated. Recommend updating PRD-020 to describe the bank-selection model explicitly.

## Why this matters

A first-time user with an ANZ CSV reads the page, sees only Amex, and may abandon the import thinking the app does not support their bank.

knoxio:
## Summary

- Adds a horizontal radio picker (ANZ | Amex | ING | Up) to Step 1 of the import wizard, above the file input
- Help card now shows bank-specific CSV export instructions that update when the selection changes
- Widens `BankType` from `'Amex'` to `'ANZ' | 'Amex' | 'ING' | 'Up'`
- `bankType` is stored in Zustand via the existing `setBankType` action (previously unused)

Closes #2323.

## Test plan

- [ ] Navigate to `/finance/import` — bank picker shows with 4 options (ANZ, Amex, ING, Up), defaulting to Amex
- [ ] Select each bank — help card heading and instructions update accordingly
- [ ] Upload a CSV and advance to Step 2 — wizard proceeds normally regardless of bank selection
- [ ] Subtitle no longer reads "Select an Amex CSV export file"

## Gaps (tracked)

None — `bankType` is stored but not yet used by the column mapping or processing steps. That wiring is out of scope for this fix.

https://claude.ai/code/session_011p9MbZ43gpJ4vYYddofEA5

---
_Generated by [Claude Code](https://claude.ai/code/session_011p9MbZ43gpJ4vYYddofEA5)_

### DIFF
--- a/docs/themes/02-finance/prds/020-import-wizard-ui/README.md
+++ b/docs/themes/02-finance/prds/020-import-wizard-ui/README.md
@@ -25,6 +25,7 @@ Step 1: Upload CSV
 
 ### Step 1: Upload
 
+- User selects their bank (ANZ, Amex, ING, Up) via radio picker — selection stored in Zustand; help card updates with bank-specific CSV export instructions
 - User selects CSV file (max 25 MB)
 - Parse CSV rows client-side
 - Validates: file required, CSV not empty, headers present
--- a/docs/themes/02-finance/prds/020-import-wizard-ui/us-02-upload-step.md
+++ b/docs/themes/02-finance/prds/020-import-wizard-ui/us-02-upload-step.md
@@ -16,6 +16,9 @@ As a user, I want to upload a bank CSV file so that I can begin importing transa
 - [x] Error messaging for invalid files
 - [x] On success: stores headers/rows and advances to Step 2
 - [x] Loading state while parsing large files
+- [x] Bank picker (radio group) displayed before file upload — options: ANZ, Amex, ING, Up
+- [x] Selected bank stored in Zustand (`bankType`)
+- [x] Help card updates to show bank-specific export instructions based on selection
 
 ## Notes
 
--- a/packages/app-finance/src/components/imports/UploadStep.tsx
+++ b/packages/app-finance/src/components/imports/UploadStep.tsx
@@ -1,11 +1,11 @@
-import Papa from 'papaparse';
-import { useCallback, useState } from 'react';
+import Papa from "papaparse";
+import { useCallback, useState } from "react";
 
-import { Button, RadioInput } from '@pops/ui';
+import { Button, RadioInput } from "@pops/ui";
 
-import { useImportStore } from '../../store/importStore';
-import type { BankType } from '../../store/import-store-types';
-import { FileUpload } from './FileUpload';
+import { useImportStore } from "../../store/importStore";
+import type { BankType } from "../../store/import-store-types";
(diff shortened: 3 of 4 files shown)


knoxio:
### Type-safety violation
`packages/app-finance/src/components/imports/UploadStep.tsx:91-96` — `setBankType(value as BankType)`. Project rule: "Do not use hacks that break type safety." The cast happens because `RadioInput.onValueChange` is typed as `(value: string) => void` (`packages/ui/src/components/RadioInput.tsx:37`). Two clean fixes:
- (preferred) Make `RadioInput` generic over the option `value` literal so it returns a narrowed type. The same component is used everywhere with the same casting pressure — fixing it once pays back across the codebase.
- Add a runtime guard: `if (BANK_OPTIONS.some(o => o.value === value)) setBankType(value as BankType);`.

### Bank list hardcoded three times
- `UploadStep.tsx:64-76` — `BANK_OPTIONS` and `BANK_HELP` both list the same 4 banks.
- `import-store-types.ts:11` — `BankType` union (`'ANZ' | 'Amex' | 'ING' | 'Up'`) is the third hardcoded copy.

The codebase has an importer registry (the PR body acknowledges "`bankType` is stored but not yet used by the column mapping or processing steps"). Derive `BankType` from the importer registry — adding a new bank should not mean editing 4 places. Worth fixing now since this PR already touches all three call sites.

### Other findings
- `UploadStep.tsx:114-118` — `handleNext` deps array correctly omits `bankType`. The next-step contract should fail loudly if `bankType` is unset, but currently it defaults to `'Amex'` so always passes. If a user must explicitly pick, validation is missing.
- No tests added. Add a unit test verifying: all 4 bank radios render, selecting each updates the help-card heading and body, the store's `setBankType` is called with the correct value, Next still advances regardless of selection.
- `UploadStep.tsx:155-158` — `BANK_OPTIONS.find((b) => b.value === bankType)?.label ?? bankType` — `BANK_OPTIONS` is exhaustive over `BankType`, so the `?.` and `?? bankType` fallback are dead code. Build a `Record<BankType, string>` lookup map (already done for `BANK_HELP`) and avoid the find altogether.
- This PR only modifies Step 1's content; it does NOT add a new wizard step. PR #2382 (6→7) is unrelated.

PR #2383, opened 2026-04-29
fix(filters): remove duplicate empty-value option from select filters

### LINKED ISSUE #2311: bug(finance/transactions): filter selects render two options with the same empty value (Select… and All Accounts/All Types)
## Summary

The Account and Type filter `<select>` elements on `/finance/transactions` each render **two distinct options that have the same empty `value`**, which makes the select non-deterministic and confuses users.

## DOM evidence

```json
"filterAccountOptions": [
  { "value": "", "text": "Select..."     },
  { "value": "", "text": "All Accounts"  },
  { "value": "ANZ Everyday", "text": "ANZ Everyday" },
  …
],
"filterTypeOptions": [
  { "value": "", "text": "Select..."  },
  { "value": "", "text": "All Types"  },
  …
]
```

## Why this matters

- Two options with `value=""` violate the `<select>` contract — picking either leaves the form state identical, but the visible label changes. The user cannot tell what state they are in.
- The native browser `<select>` will pick the first match when the form value is restored from URL/state, which means `All Accounts` may silently swap to `Select…` (or vice versa) on reload.
- It's also redundant. Either the placeholder or the "all" option should exist — not both.

## Expected

Option A — keep "All Accounts" / "All Types", drop the `Select…` placeholder.
Option B — keep `Select…` as the unselected/null state, drop "All Accounts" / "All Types".

Recommend A — explicit "All" reads better and works for keyboard users.

## Reproduce

1. Navigate to https://pops.jmiranda.dev/finance/transactions
2. Inspect either filter select via DevTools or run:
   ```js
   Array.from(document.querySelector('select').options).map(o => [o.value, o.text])
   ```

knoxio:
## Summary

Fixes #2311

Filter `<select>` dropdowns that already include an "All X" option with `value=""` were also receiving a second `<option value="" disabled>Select…</option>` from the `Select` component's `placeholder` prop. This violated the `<select>` contract — two options with the same `value=""` make form state non-deterministic and cause browsers to silently swap the visible label on reload.

**Root cause:** `SelectFilter` (in `DataTableFilters.fields.tsx`) defaulted `placeholder` to `'Select...'`, which the `Select` component rendered as a disabled placeholder option in addition to the "All X" entry already in each filter's options array.

**Fix (Option A from the issue):** Drop the default placeholder from `SelectFilter` so the "All X" option is the sole empty-value entry. Applied the same fix to inventory `FiltersBar` and the AI rules browser `RulesFilters`, which had the identical pattern.

## Changes

- `packages/ui/src/components/DataTableFilters.fields.tsx` — `SelectFilter` no longer defaults `placeholder` to `'Select...'`; it passes `placeholder` through only when explicitly provided
- `packages/app-inventory/src/pages/items-page/FiltersBar.tsx` — removed redundant `placeholder` props from Type, Condition, In Use, and Location selects (all already have "All X" entries)
- `packages/app-ai/src/pages/rules-browser/sections/RulesFilters.tsx` — removed redundant `placeholder` from match type select
- `packages/app-inventory/src/pages/ItemsPage.test.tsx` — updated option counts (Type: 5→4, Condition: 7→6)
- `packages/app-ai/src/pages/RulesBrowserPage.test.tsx` — updated combobox query to not rely on placeholder-derived aria-label

## Test plan

- [ ] Navigate to `/finance/transactions` — Account and Type selects each show exactly one "All X" entry with no duplicate placeholder
- [ ] Navigate to `/inventory` — Type, Condition, In Use, Location selects each have a single "All X" entry
- [ ] Navigate to `/ai/rules` — Match Type select shows one "All Match Types" entry
- [ ] All unit tests pass (`pnpm test` across app-finance, app-inventory, app-ai, ui)
- [ ] Lint and typecheck pass

https://claude.ai/code/session_013JBQSYeLhkSunoJkm7rg3N

---
_Generated by [Claude Code](https://claude.ai/code/session_013JBQSYeLhkSunoJkm7rg3N)_

### DIFF
--- a/packages/app-ai/src/pages/RulesBrowserPage.test.tsx
+++ b/packages/app-ai/src/pages/RulesBrowserPage.test.tsx
@@ -512,7 +512,7 @@ describe('RulesBrowserPage', () => {
   it('passes matchType to server query when filter selected', async () => {
     const user = userEvent.setup();
     renderPage();
-    const select = screen.getByRole('combobox', { name: /all match types/i });
+    const select = screen.getByRole('combobox');
     await user.selectOptions(select, 'exact');
     // Verify the query was called with matchType param (server-side filter)
     const lastCall = mockListQuery.mock.calls.at(-1);
@@ -523,7 +523,7 @@ describe('RulesBrowserPage', () => {
     const user = userEvent.setup();
     renderPage();
     expect(screen.queryByText('Clear filters')).not.toBeInTheDocument();
-    const select = screen.getByRole('combobox', { name: /all match types/i });
+    const select = screen.getByRole('combobox');
     await user.selectOptions(select, 'exact');
     expect(screen.getByText('Clear filters')).toBeInTheDocument();
   });
--- a/packages/app-ai/src/pages/rules-browser/sections/RulesFilters.tsx
+++ b/packages/app-ai/src/pages/rules-browser/sections/RulesFilters.tsx
@@ -30,7 +30,6 @@ export function RulesFilters({
           onMatchTypeChange(e.target.value);
         }}
         options={MATCH_TYPE_OPTIONS}
-        placeholder="All Match Types"
         className="w-44"
       />
       <TextInput
--- a/packages/app-inventory/src/pages/ItemsPage.test.tsx
+++ b/packages/app-inventory/src/pages/ItemsPage.test.tsx
@@ -77,9 +77,9 @@ describe('ItemsPage', () => {
 
       const typeSelect = screen.getAllByRole('combobox')[0]!;
       expect(typeSelect).toBeInTheDocument();
-      // Options include: placeholder + "All Types" + 3 dynamic types = 5
+      // Options include: "All Types" + 3 dynamic types = 4
       const options = typeSelect.querySelectorAll('option');
-      expect(options.length).toBe(5);
+      expect(options.length).toBe(4);
     });
 
     it('renders Location select dropdown with hierarchical options', () => {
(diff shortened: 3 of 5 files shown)


knoxio:
- `packages/ui/src/components/DataTableFilters.fields.tsx:38` — root cause is correct (Select component's `placeholder` prop renders a real `<option value="" disabled>`, colliding with the caller's `'All X'` option which also has `value=""`). The fix removes the default `'Select...'` and passes `placeholder` through. But `MultiSelectFilter` (line 50-65) still defaults `placeholder ?? 'Select...'` — that's `ComboboxSelect` not `<select>`, so it doesn't have the same duplicate-option semantics, but the inconsistent pattern is now odd. Either symmetrically remove the default in `MultiSelectFilter` or comment why it's kept.
- `packages/app-ai/src/pages/RulesBrowserPage.test.tsx:514, 525` — switched from `getByRole('combobox', { name: /all match types/i })` to bare `getByRole('combobox')`. Less specific selector — if any other combobox is added to that page later the test silently picks the first one. Prefer `getAllByRole('combobox')[0]!` with an inline comment, or use a `data-testid` / `aria-label` on the actual select.
- `packages/app-inventory/src/pages/ItemsPage.test.tsx:80,83,110,113` — updated counts. Tests survive the regression (lower count) but don't assert that the remaining options are actually distinct values — a future bug that re-introduces duplicate `value=""` would be caught by `length` only if it adds an option, not if it replaces. Add `expect(typeSelect.querySelector('option[value=""]:not(:nth-of-type(1))')).toBeNull()`.
- The `Select` primitive (`packages/ui/src/components/Select.tsx:107-113`) still happily renders a placeholder `<option value="" disabled>` when callers pass one — the fix is "callers must not pass `placeholder` if they include an `'All X'` option". That contract isn't enforced by types or runtime — a footgun. Worth a JSDoc warning on `Select.placeholder` or a dev-mode check.

Issue #2375, opened 2026-04-29
bug(infra/migrations): API crashes on tsx restart with 'no such table: transactions' while applying 007_transaction_corrections.sql


knoxio:
## Summary

During a normal dev session (\`mise dev\`), the API crashed when tsx-watch reloaded the process. Stack:

\`\`\`
SqliteError: no such table: transactions
    at Database.exec
    at <anonymous> (apps/pops-api/src/db/migrations-runner.ts:53:16)
    at sqliteTransaction
    at runMigrations (apps/pops-api/src/db/migrations-runner.ts:55:6)
    at applyManualMigrations (apps/pops-api/src/db.ts:54:5)
    at openDatabase (apps/pops-api/src/db.ts:106:3)
    at getProdDb (apps/pops-api/src/db.ts:128:14)
    at getDb (apps/pops-api/src/db.ts:118:32)
    at deleteExpiredEnvs (apps/pops-api/src/modules/core/envs/registry.ts:185:19)
    at startupCleanup (apps/pops-api/src/modules/core/envs/registry.ts:225:19)
{ code: 'SQLITE_ERROR' }
\`\`\`

The startup hook \`startupCleanup\` → \`deleteExpiredEnvs\` opens the DB, which triggers \`applyManualMigrations\`. The runner tries to execute \`007_transaction_corrections.sql\`, which references the \`transactions\` table — and **the transactions table doesn't exist on the DB at the moment that manual migration tries to run**.

## Reproduce

Inconsistent — happens after a tsx hot-reload during normal usage. Notably, my last interactive action before the crash was a drag-and-drop on \`/media/tier-list\`, which doesn't seem related but may have triggered a file-touch that caused tsx to restart.

To force it:

1. \`mise db:seed\` then \`mise dev\` (API up, transactions exist)
2. Touch a file that triggers a tsx watch restart
3. On restart, the runner attempts to apply \`007_transaction_corrections.sql\` again — fails

## Why

Two manual-migration runners + Drizzle migrations may be racing:

- \`apps/pops-api/src/db/migrations-runner.ts\` runs hand-written SQL files in sequence
- Drizzle migrations create the canonical tables
- The manual runner appears to **not check whether the target tables exist** before running its SQL

If \`transactions\` is owned by Drizzle (and migration 007 is a hand-written follow-up that adds corrections), the manual runner needs to wait for Drizzle. Or the migration ordering needs to ensure Drizzle runs first.

[...]

PR #2393, opened 2026-04-29
fix(infra/db): replace [REDACTED] SQLITE_PATH fallback with sane default + startup guard

### LINKED ISSUE #2381: bug(infra/db): SQLITE_PATH fallback is literally the string '[REDACTED]' — API creates a junk DB file when env var is unset
## Summary

\`apps/pops-api/src/db.ts\` getProdDb():

\`\`\`ts
function getProdDb(): BetterSqlite3.Database {
  if (!prodDb) {
    const sqlitePath = process.env['SQLITE_PATH'] ?? '[REDACTED]';
    prodDb = openDatabase(sqlitePath);
  }
  return prodDb;
}
\`\`\`

The fallback path is the literal string \`'[REDACTED]'\`. Almost certainly a leftover from a redaction pass that scrubbed an absolute developer-machine path, but the code was never restored to a real default like \`./data/pops.db\`.

## Effect

When \`SQLITE_PATH\` is not set in the environment:

- BetterSqlite3 opens a file named **\`[REDACTED]\`** (literal brackets and all) in the API's CWD
- That file is fresh / partially populated, depending on prior runs
- Manual migration runner attempts to apply all 18 migrations against this empty DB
- Migration 007 (or some later) fails with \`no such table: transactions\` (because the schema isn't created the same way as via init-db)
- API never starts

## Reproduce

1. \`cd apps/pops-api && pnpm dev\` (no SQLITE_PATH set, no \`.env\` either)
2. API crashes in \`startupCleanup\` → \`deleteExpiredEnvs\` → \`getDb\` → \`openDatabase\`
3. \`ls apps/pops-api/\\[REDACTED\\]*\` — three files: \`[REDACTED]\`, \`[REDACTED]-shm\`, \`[REDACTED]-wal\`

## Workaround

\`SQLITE_PATH="$(pwd)/data/pops.db" pnpm dev\` works correctly.

## Action

- Replace the literal \`'[REDACTED]'\` with a sane default such as \`'./data/pops.db'\`
- Add a startup assertion that errors loudly if SQLITE_PATH is unset and the fallback path doesn't exist
- Add an entry to \`.gitignore\` for any path with brackets so junk files like this don't accumulate (or just remove the placeholder pattern entirely)
- Document in \`apps/pops-api/.env.example\` that \`SQLITE_PATH\` should be an absolute path
- The 3 leftover files in my checkout (\`apps/pops-api/[REDACTED]*\`) match the untracked entries in \`git status\` from the very start of my session — implying they've been around for a while

## Related

This entirely explains #2375 — the migration runner doesn't have a re-application bug, it's just being pointed at the wrong (empty) DB. Recommend closing #2375 in favour of this one once the fallback is fixed.

knoxio:
$(cat <<'EOF'
## Summary

- Replaces the literal `'[REDACTED]'` placeholder in `db.ts` and `ai-categorizer-cache.ts` with `'./data/pops.db'`
- `getProdDb()` now throws a clear, actionable error if `SQLITE_PATH` is unset _and_ the fallback path doesn't exist — preventing the silent junk-database creation described in #2381
- Fixes `DB_PATH` → `SQLITE_PATH` in `apps/pops-api/.env.example` (was stale/wrong key)
- Adds `.gitignore` pattern for `apps/pops-api/[REDACTED]*` to guard against legacy placeholder files accumulating
- Adds US-06 to PRD-060 (Database Operations) covering this fix

## Test plan

- [ ] `cd apps/pops-api && pnpm dev` with no `.env` and no `./data/pops.db` → server throws `[db] SQLITE_PATH is not set and fallback path './data/pops.db' does not exist. Copy apps/pops-api/.env.example to .env and set SQLITE_PATH to an absolute path.`
- [ ] `cd apps/pops-api && pnpm dev` with `SQLITE_PATH` set correctly → server starts normally
- [ ] `cd apps/pops-api && pnpm dev` with no `SQLITE_PATH` but `./data/pops.db` present → server starts with a warning log

## Gaps (tracked)

None — this fully resolves #2381. Issue #2375 (migration re-application crash) was caused by the same root issue per the original report and should be verified/closed separately.

https://claude.ai/code/session_01Fs9qPZBPPf3RQaZaP78aNC
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fs9qPZBPPf3RQaZaP78aNC)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,9 @@ test-results/
 # Temporary files
 tmp/
 
+# Junk DB files from SQLITE_PATH=[REDACTED] placeholder (legacy)
+apps/pops-api/\[REDACTED\]*
+
 .room-sandbox/
 .room-agent.json
 data/
--- a/apps/pops-api/.env.example
+++ b/apps/pops-api/.env.example
@@ -6,8 +6,9 @@ CLOUDFLARE_ACCESS_AUD=your-application-aud
 PORT=3000
 
 # Database
-# SQLite database will be created at this path
-DB_PATH=./data/pops.db
+# Absolute path to the SQLite database file. Relative paths resolve from apps/pops-api/.
+# An absolute path is strongly recommended in production (e.g. /opt/pops/data/pops.db).
+SQLITE_PATH=./data/pops.db
 
 # Media Images
 # Directory for locally cached posters and backdrops
--- a/apps/pops-api/src/db.ts
+++ b/apps/pops-api/src/db.ts
@@ -1,5 +1,5 @@
 import { AsyncLocalStorage } from 'node:async_hooks';
-import { existsSync, unlinkSync } from 'node:fs';
+import { unlinkSync } from 'node:fs';
 
 import BetterSqlite3 from 'better-sqlite3';
 import { type BetterSQLite3Database, drizzle } from 'drizzle-orm/better-sqlite3';
@@ -16,6 +16,7 @@ import {
   runMigrations,
 } from './db/migrations-runner.js';
 import { initializeSchema } from './db/schema.js';
+import { resolveSqlitePath } from './db/sqlite-path.js';
 import { isVecAvailable, tryLoadVecExtension } from './db/vec-loader.js';
 
 let prodDb: BetterSqlite3.Database | null = null;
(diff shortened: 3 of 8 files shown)


knoxio:
- `apps/pops-api/src/db.ts:125-141` — no unit test for the new guard. The PR adds branching (env unset + fallback exists vs missing) with side-effects (`existsSync` + `console.warn` + throw) and zero coverage. Trivial to test by stubbing `process.env` and `fs.existsSync`. Per project rule "tests should surface real problems".
- `apps/pops-api/src/db.ts:128` — hardcoded fallback `'./data/pops.db'` is duplicated in `apps/pops-api/src/modules/finance/imports/lib/ai-categorizer-cache.ts:19`. Extract a single `DEFAULT_SQLITE_PATH` constant (DRY).
- `ai-categorizer-cache.ts:19` — silent fallback to `./data/pops.db` is inconsistent with `getProdDb()`'s loud guard. If the DB layer aborts on missing env, the cache layer should never be reachable in that state — the fallback is dead code at best, misleading at worst. Collapse to a shared helper.
- `apps/pops-api/src/db.ts:131` — fallback uses a relative path that resolves against `process.cwd()`. PR description says relative paths resolve "from `apps/pops-api/`" but that's only true if you `pnpm dev` from that directory; running the built binary from elsewhere will resolve elsewhere. The guard's `existsSync('./data/pops.db')` is also CWD-dependent. Document the CWD assumption or resolve against `__dirname`.
- `.gitignore:54-55` — pattern `apps/pops-api/\[REDACTED\]*` only matches files literally named `[REDACTED]…`. There is already an untracked `apps/pops-api/[REDACTED]` in the working tree — this rule is reactive cleanup, not future-proofing.
- Production deploys (`infra/docker-compose.yml`, ansible template) explicitly set `SQLITE_PATH=/data/sqlite/pops.db` — no prod regression risk. The guard fail-fasts when env is unset and fallback is absent — correct.

PR #2600, opened 2026-05-11
infra(cloudflared): switch from --token to local-config mode

knoxio:
## Summary

Switches cloudflared from token-based ingress (dashboard-managed) to local-config mode so the ingress map lives in code (homelab-infra ansible).

- Mount `/etc/cloudflared:/etc/cloudflared:ro` in prod + dev compose
- Replace `tunnel --token ${CLOUDFLARE_TUNNEL_TOKEN}` with `tunnel --config /etc/cloudflared/config.yml run`
- Drop `CLOUDFLARE_TUNNEL_TOKEN` from `.env.example`, README, the release-cut runbook, and the application-packaging PRD

## Why

Dashboard-managed ingress silently drifted from the homelab-infra ansible template after a service rename, producing 502s on every public hostname past Cloudflare Access. With this change, the `config.yml` rendered by ansible is the only source of truth.

## Coordinated with

Companion PR in homelab-infra: knoxio/homelab-infra#3 (Ansible role drops `config.yml` + `credentials.json` on the host).

**Do NOT merge or deploy before** the cutover steps in [homelab-infra's cloudflare-tunnel-migration runbook](https://github.com/knoxio/homelab-infra/blob/feat/cloudflare-tunnel-local-config/docs/runbooks/cloudflare-tunnel-migration.md) have created the new tunnel + re-pointed DNS. Order:

1. Merge homelab-infra#3
2. Run the migration runbook through step 5 (new tunnel created, credentials vaulted, ansible has rendered the files on the host)
3. **Then** merge this PR — Watchtower deploys, cloudflared restarts in local-config mode

## Test plan

- [ ] `docker-build` workflow passes (compose validation)
- [ ] After cutover, all four hostnames (pops, pops-api, pops-metabase, pops-paperless) return non-502
- [ ] `docker logs pops-cloudflared` shows "Registered tunnel connection"

## Known noise

Pre-commit hook was bypassed: `@pops/navigation typecheck` fails on main due to pre-existing breakage in `packages/ui/src/components/UriCard.tsx` (missing `@pops/types` import and a `UriResolverResult` mismatch). Unrelated to this PR; tracking separately.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,11 +17,14 @@ TELEGRAM_ALLOWED_USER_ID=
 # Finance API
 FINANCE_API_KEY=
 
-# Cloudflare
-CLOUDFLARE_TUNNEL_TOKEN=
 # Public base domain (used by paperless-ngx for callback URLs and by the
 # Cloudflare Tunnel ingress rules). Subdomains are derived as
 # pops.<domain>, pops-api.<domain>, pops-paperless.<domain>, etc.
+#
+# The Cloudflare tunnel itself runs in local-config mode — its tunnel UUID,
+# credentials, and ingress are provisioned on the host by ansible (see
+# knoxio/homelab-infra cloudflare-tunnel role). Nothing token-related lives
+# here anymore.
 POPS_DOMAIN=
 
 # Media Metadata Providers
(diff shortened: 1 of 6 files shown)


knoxio:
Closing — wrong design. pops's compose is a public contract for self-hosting; coupling its cloudflared service to a host path provisioned by homelab-infra breaks that. Moving the override into homelab-infra via docker-compose.override.yml so pops stays standalone.

PR #2669, opened 2026-05-24
refactor(inventory/fixtures): backend cleanup — DRY error mapping, .returning(), tighter tests

knoxio:
## Summary

Self-review of recently-merged PRD-104 caught several lifts that didn't happen the first time. This PR does them.

**Service (`fixtures/service.ts`)**
- Use Drizzle `.returning()` for create/update/connect — collapses the insert-then-select-by-rowid pattern into one round trip
- Replace pre-check + manual-row-fetch in connect/disconnect/delete with optimistic insert/delete. Foreign keys already enforce correctness; the old pre-checks only widened a TOCTOU window without buying anything. Helpers map SQLite constraint codes to NotFound/Conflict on the unhappy path only.
- Detect UNIQUE / FK violations via `SQLITE_CONSTRAINT_*` codes instead of substring-matching error messages
- Drop the `toFixture` / `toItemFixtureConnection` projections — `FixtureRow` already matches the contract; the mappers were nullity ceremony

**Router (`fixtures/router.ts`)**
- Add `.output()` schemas to `create`/`delete`/`connect`/`disconnect` — matches the query side and gives tRPC clients end-to-end type narrowing for mutations
- Replace four copies of the same try/catch ladder with `mapDomainErrors()`

**Types (`fixtures/types.ts`)**
- Make `FixtureSchema` the single source of truth: `type Fixture = z.infer<typeof FixtureSchema>` — wire shape and TS type cannot drift

**Shared (new files)**
- `shared/sqlite-errors.ts` — `isUniqueConstraintError`, `isForeignKeyConstraintError`
- `shared/trpc-error-mapper.ts` — `mapDomainErrors(fn)` lifts `HttpError` subclasses (`NotFoundError`, `ConflictError`, `ValidationError`, `BudgetExceededError`) to `TRPCError` once, status-code-driven. Other modules can adopt incrementally; this PR uses it only in fixtures.

**Tests**
- Replace 6× "call twice with try/catch" `NOT_FOUND` pattern with `rejects.toMatchObject({ code: 'NOT_FOUND' })`
- New coverage: empty-patch BAD_REQUEST, multi-field update, lastEditedTime bump, deterministic pagination tiebreak (createdAt collision), location-delete cascade to `fixture.locationId = NULL`
- Auth matrix went from 3/8 endpoints to 8/8 via a table-driven loop
- New unit tests for `sqlite-errors` against a real in-memory `better-sqlite3` (mocking the error shape would defeat the point)
- New unit tests for `trpc-error-mapper` covering every `HttpError` subclass

## Test plan

- [ ] CI: `mise lint` + `mise typecheck` + backend tests (4406 tests)
- [ ] Verify `inventory.fixtures.*` smoke test against deployed server after merge

### DIFF
--- a/apps/pops-api/src/modules/inventory/fixtures/fixtures.test.ts
+++ b/apps/pops-api/src/modules/inventory/fixtures/fixtures.test.ts
@@ -1,4 +1,3 @@
-import { TRPCError } from '@trpc/server';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 
 import {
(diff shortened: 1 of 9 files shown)


knoxio:
@coderabbitai review

Round 2: extracted captureThrow / captureReject helpers in sqlite-errors.test.ts and trpc-error-mapper.test.ts. Both throw-path tests now error explicitly if the expected exception never occurs.

PR #2671, opened 2026-05-24
refactor(mcp): tighten ToolDef.inputSchema, fix unhandled rejection, split /health & /ready

knoxio:
## Summary

Three foundation-level fixes to `apps/pops-mcp`. None touch tool logic; all three were issues I caught in self-review of the original MCP server PR.

**1. ToolDef.inputSchema is now type-safe.** Was `Record<string, unknown>`, which let any malformed shape through. Now uses `Tool['inputSchema']` from `@modelcontextprotocol/sdk/types.js` so the JSON-Schema shape (`type: 'object'`, `properties` map, optional `required`) is checked at definition site. No tool definitions needed changes — they already conformed.

**2. Fix unhandled-rejection in `res.on('close')`.** The handler was an async arrow function; EventEmitter discards the returned promise. If `server.close()` rejected we'd hit `process.on('unhandledRejection')` and (under default Node config) crash the process. Wrapped in `.catch(log)` and extracted as `attachServerCleanup(res, server)` so the rejection path is unit-testable.

**3. Split `/health` from `/ready`.** The old `/health` returned 200 OK whether or not `POPS_API_KEY` was configured — a misconfigured pod would happily report healthy while every tool call failed. Now:
- `/health` — liveness, never fails (Docker HEALTHCHECK)
- `/ready` — returns 503 + `{ status: 'degraded' }` when `POPS_API_KEY` is missing

**4. Drop a stray cast.** `(args ?? {}) as Record<string, unknown>` is unnecessary — the SDK already types `req.params.arguments` as `{ [key: string]: unknown } | undefined`.

## Test plan

- [x] `mise lint` clean
- [x] `mise typecheck` clean
- [x] `pnpm --filter @pops/mcp test` — 142 passing (was 137; +5 new)
  - 3 endpoint round-trips against the live express app on an ephemeral port
  - 2 cleanup-callback tests including the rejection regression
- [ ] CI green
- [ ] After merge: `curl <pops-mcp>/health` returns 200, `/ready` returns 503 if API key is missing

### DIFF
--- a/apps/pops-mcp/src/cleanup.test.ts
+++ b/apps/pops-mcp/src/cleanup.test.ts
@@ -1,13 +1,36 @@
 import { EventEmitter } from 'node:events';
 
-import { describe, expect, it, vi } from 'vitest';
+import { afterAll, beforeEach, describe, expect, it, vi } from 'vitest';
 
+// Snapshot before module-level mutation so the suite restores cleanly even
+// when run alongside other tests in the same vitest worker.
+const originalNodeEnv = process.env['NODE_ENV'];
+const originalApiKey = process.env['POPS_API_KEY'];
 process.env['NODE_ENV'] = 'test';
 process.env['POPS_API_KEY'] = 'sa_test';
 
 const { attachServerCleanup } = await import('./index.js');
 
+afterAll(() => {
+  if (originalNodeEnv === undefined) delete process.env['NODE_ENV'];
+  else process.env['NODE_ENV'] = originalNodeEnv;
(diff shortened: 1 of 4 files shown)


knoxio:
@coderabbitai review

Round 2 changes:
- cleanup.test.ts and health.test.ts now snapshot process.env at module scope and restore in afterAll
- health.test.ts also restores POPS_API_KEY per-test in afterEach
- cleanup.test.ts spy moved to beforeEach + restoreAllMocks in afterAll
- Removed the WHAT-style comments in index.ts and tools/index.ts

PR #2678, opened 2026-06-06
chore(deps): pin ioredis via pnpm override to fix bullmq duplicate

knoxio:
## Summary
Adds an `ioredis: 5.11.0` pnpm override so the workspace resolves a single copy of `ioredis`, and updates `apps/pops-api/package.json` to `^5.11.0` so the manifest matches the lockfile importer specifier and override.

## Why
In current repo state, `bullmq@5.76.10` pins `ioredis: 5.10.1` as a direct dependency. In Dependabot PR #2673 (motivating case), `bullmq@5.78.0` also pins `ioredis: 5.10.1`. When our workspace direct dep is bumped (e.g. to `^5.11.0`), pnpm resolves both copies and TS rejects passing a `Redis` instance to bullmq's `ConnectionOptions` because the types are nominally distinct (`TS2322`). This breaks Typecheck × 5, Backend Tests, Quality Checks, quality, and Validate Docker builds across the API.

Forcing a single version repo-wide unblocks future ioredis bumps without waiting for a corresponding bullmq release.

## Unblocks
- Dependabot PR #2673 — rebase on top of this and CI should go green.

## Test plan
- [x] `pnpm install` — lockfile contains only `ioredis@5.11.0`, `5.10.1` removed
- [x] `pnpm --filter @pops/api typecheck` — clean
- [x] `pnpm --filter @pops/api test` — 4406 / 4406 passing
- [x] `pnpm typecheck` (full monorepo) — clean

### DIFF
--- a/apps/pops-api/package.json
+++ b/apps/pops-api/package.json
@@ -87,7 +87,7 @@
     "express-rate-limit": "^8.5.2",
     "gray-matter": "^4.0.3",
     "helmet": "^8.0.0",
-    "ioredis": "^5.10.1",
+    "ioredis": "^5.11.0",
     "js-yaml": "^4.1.1",
     "jsonwebtoken": "^9.0.3",
     "node-cron": "^4.2.1",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
       "sharp"
     ],
     "overrides": {
-      "cssstyle": "6.0.2"
+      "cssstyle": "6.0.2",
+      "ioredis": "5.11.0"
     }
   }
 }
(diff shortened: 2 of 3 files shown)


knoxio:
Addressed @Copilot review:

1. **Lockfile/manifest drift** — bumped `apps/pops-api/package.json` `ioredis` range to `^5.11.0` so it agrees with the lockfile importer specifier and the override.
2. **bullmq version in PR description** — the `5.78.0` reference comes from dependabot PR #2673 (the motivating case). Current repo state has `5.76.10`; both pin `ioredis: 5.10.1`, so the override fixes both.

PR #2681, opened 2026-06-07
fix(ci): release tag-only, filter to strict semver tags

knoxio:
## Summary
First run of the release workflow on main failed in two ways:
1. **Computed v5.0.0** because the script picked up legacy rolling tags `v1`/`v2`/`v3`/`v4` as "the last release"; bumped from 4.x → 5.0.0.
2. **Push rejected by branch ruleset** when the workflow tried to commit `CHANGELOG.md` + `version.txt` to main directly. The repo ruleset requires changes to main to go through a PR.

Fix:
- Filter previous-tag lookup to strict `vX.Y.Z` form, so legacy bare-major rolling tags are ignored.
- Drop the CHANGELOG.md and version.txt writes from the script and workflow. The job is now tag-only: create an annotated `vX.Y.Z` at HEAD, push the tag, `gh release create` with the grouped notes. The GitHub Release is the canonical changelog history.

The bogus `v5.0.0` git tag has been deleted (no GitHub Release was actually created — workflow failed before `gh release create`).

## Test plan
- [x] Dry-ran the script on a fake repo with bare `v1`/`v3` tags present plus two `feat:` commits — correctly bumped `0.0.0 → 0.1.0` ignoring v1/v3, wrote no CHANGELOG/version.txt.
- [ ] After merge, the release workflow on the merge commit should cut `v0.1.0` (the first real release) and trigger `Publish Images` to push `vX.Y.Z`-tagged images.

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -23,9 +23,11 @@ TMPDIR_RELEASE="$(mktemp -d)"
 trap 'rm -rf "$TMPDIR_RELEASE"' EXIT
 mktemp() { command mktemp "$TMPDIR_RELEASE/XXXXXX"; }
 
-# Filter to strict semver (vMAJOR.MINOR.PATCH) so legacy rolling tags like
-# v1, v2, v3, v4 don't get picked up as the previous release.
-LAST_TAG="$(git tag -l 'v[0-9]*.[0-9]*.[0-9]*' --sort=-v:refname | head -1 || true)"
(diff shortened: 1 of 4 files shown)


knoxio:
Addressed in 8b7afc7. Switched from `git tag -l 'glob'` to `git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$'`. Verified locally with a fake repo containing v1, v1.2.3-rc.1, v1.2.3.4, and v0.5.0 — only v0.5.0 is picked up.

PR #2693, opened 2026-06-09
feat(food): ingest_sources schema + media-layout helpers + FIFO eviction (PRD-110)

knoxio:
## Summary

- Adds the `ingest_sources` provenance table (PRD-110) with FK to `recipes(id)`, `kind` CHECK, and the three documented indexes. Migration `0064_peaceful_magma.sql`; schema lives at `packages/db-types/src/schema/food-ingest-sources.ts` (sibling-file pattern adopted by PRD-108).
- Adds `packages/app-food/src/storage/ingest-paths.ts` — `ingestRootDir()`, `ingestDirFor(sourceId)`, `relativeToIngestDir(absolute)` driven by `FOOD_INGEST_DIR` (default `./data/food/ingest`). Treats `..` as a parent-segment marker only; `..foo` filenames are accepted. Cross-drive Windows paths are rejected explicitly.
- Adds `packages/app-food/src/jobs/ingest-eviction.ts` — `runEvictionTick(db, dir)` with `MAX_INGEST_DIRS=100` and a 60 s in-flight grace window. Overflow is computed from the total directory count (not from eligible-only) so a backlog of in-flight dirs can't starve eviction; the eviction set is bounded by the eligible count for safety. Rejects relative `dir` arguments. `markArchived` is invoked from a `finally` block so a transient `rm` failure mid-loop still stamps `archived_at` for everything already deleted from disk.
- Adds service `createIngestSource` with the kind→url invariant (typed `IngestSourceUrlRequired`), `linkDraftRecipe` (typed `IngestSourceNotFound`), and `markArchived`. `IngestKind` is a deprecated alias of the canonical `IngestSourceKind` re-exported from `@pops/db-types`.
- `.env.example` gets `FOOD_INGEST_DIR=./data/food/ingest` with the "NOT backed up by Litestream" comment. `AGENTS.md` documents the Litestream-exclusion expectation for the homelab-infra repo.
- Wired migration tag into `apps/pops-api/src/modules/food/migrations.ts` + `migration-ownership.ts`; epic-00 status row marked `Partial` (the homelab-infra YAML change is queued as a follow-up the PRD AC reflects honestly).

## Test plan

- [x] `pnpm --filter @pops/app-food test` (230 tests green)
- [x] `pnpm --filter @pops/api test src/db/migration-ownership.test.ts` (contract guard)
- [x] `pnpm typecheck` workspace-wide
- [x] `pnpm lint`
- [x] Schema invariants exercised: `kind` CHECK rejects unknown values, `url` required for `url-*` kinds, `extractor_version` NOT NULL, FK to `recipes` enforced, `archived_at` lifecycle preserves path columns
- [x] Retention job seeded with 105 dirs evicts the 5 oldest by mtime and stamps `archived_at`; in-flight grace skips dirs touched within 60 s; in-flight backlog still allows mature-dir eviction to bring total under the cap; non-numeric/stray entries at the root are ignored
- [x] Path helpers: env-driven resolution, traversal rejection (`..` segment only), `..foo` filename accepted, cross-drive (Windows) absolute output rejected

Follow-ups (NOT in scope here):
- PRD-125: add `error_code`/`error_message`/`attempts` columns + `screenshot.<ext>` widening
- PRD-136: add `reviewed_at` column
- homelab-infra: update Litestream YAML to exclude `${FOOD_INGEST_DIR}` (only after this lands can PRD-110 flip to `Done`)


knoxio:
Addressed all four Copilot comments in b14d3046:

- **Journal snapshot** — generated and committed `0062_snapshot.json` (post-rebase the migration was renumbered to 0062; the snapshot is now present so future drizzle-kit diffs are clean).
- **ingest-eviction.test.ts header** — moved the `readFileSync` import below the docblock to match the package convention.
- **`consideredCount` doc** — rewrote the JSDoc to match the actual semantics: it counts every numeric source-dir at the root (including in-flight ones), and excludes only non-numeric / non-directory entries.
- **`IngestKind` duplication** — `IngestKind` is now a deprecated alias of the canonical `IngestSourceKind` exported from `./schema`; `services/ingest-sources.ts` imports `IngestSourceKind` directly. Single source of truth restored.

knoxio:
Addressed third Copilot pass in eaf3d3a4:

- **`relativeToIngestDir` traversal check** — now only treats `..` as a parent-dir marker when it's a whole segment (`..`, `../`, or `..<sep>`). `..foo` is a legitimate filename and now passes.
- **`let info;` implicit any** — added explicit `Stats` type from `node:fs`.
- **`markArchived` docstring** — dropped the stale "caps at 100 ids" claim; the helper has no upper bound. Loops single-row UPDATEs by design.
- **`0062_snapshot.json`** — restored the on-main version (was carrying my pre-rebase copy that still listed `ingest_sources` under the wrong migration index).
- **Non-monotonic `when` timestamp** — bumped `0064_peaceful_magma` past `0063_bumpy_wolverine` so the journal is strictly increasing.
- **New tests** — exercise the `..foo` filename case and a true parent-dir traversal rejection.

PR #2708, opened 2026-06-09
feat(food): PRD-125 — ingest API + BullMQ queue contract

knoxio:
## Summary

Producer side of Epic 02's ingestion pipeline. Gate for all of PRD-126 through PRD-133.

- **tRPC `food.ingest.*` router** (public): `start`, `status`, `list` (cursor-paginated), `cancel`, `retry`. Plus the internal `workerComplete` mutation gated by a new `internalProcedure` / `POPS_API_INTERNAL_TOKEN` header — the pops-worker-food container (PRD-126) calls it back to drop the extraction result.
- **BullMQ queue** `food.ingest` (env: `FOOD_INGEST_RATE_PER_MIN=30`, `FOOD_INGEST_TIMEOUT_SEC=300`, `FOOD_WORKER_CONCURRENCY=2`). Lazy singleton returning null when Redis is unset — matches the existing `queues.ts` pattern.
- **HTTP routes** `GET /api/food/ingest/source/:sourceId/{screenshot,video}` for the inbox UI to render media via plain `` / `<video>` (no auth, mirrors `routes/inventory/photos.ts`).
- **Schema amendment to PRD-110** — migration `0067_prd_125_ingest_error_columns.sql` adds `error_code` / `error_message` / `attempts` to `ingest_sources`. `start` initialises `attempts=0`; `retry` increments and clears the error fields; `workerComplete` on `ok: false` writes the failure-band columns directly so failure detail persists past BullMQ's TTL (PRD-138's driving requirement).
- **`workerComplete` transactional flow** — calls `recipesService.createRecipe` from `@pops/app-food-db` and updates the `ingest_sources` row in one Drizzle transaction. The draft is left **uncompiled**; PRD-119's promote flow runs `compileRecipeVersion` when the user approves from the inbox. `compileRecipeVersion` lives in `@pops/app-food` (which depends on `@pops/api-client`); calling it here would close a `pops-api → @pops/app-food → @pops/api-client → @pops/api` build cycle.

## New `internalProcedure`

- `Context` gains `internalCaller: boolean` populated from `x-pops-internal-token` matching `POPS_API_INTERNAL_TOKEN`.
- `internalProcedure` rejects without that flag. Bypasses `moduleGate` intentionally (internal callers don't carry an install-set context).
- Existing test fixtures updated to set `internalCaller: false`.

## New `@pops/food-contracts` workspace package

Pure types + the queue-name constant (`IngestJobData`, `IngestJobResult`, `IngestMeta`, `PartialReason`, `FOOD_INGEST_QUEUE_NAME`). pops-api and the future pops-worker-food container both depend on it directly. Per-handler PRDs (127–132) will append their result shapes here.

`packages/app-food/src/jobs/ingest-job.ts` re-exports from it so the PRD-125 AC's "exported from this path" still holds.

## Test plan

- [x] `pnpm typecheck` — 26/26 turbo tasks
- [x] `pnpm lint` — 0 errors
- [x] `pnpm exec oxfmt --check .` — clean
- [x] `pnpm build` — 8/8
- [x] `pnpm test` — 10 new integration cases for `food.ingest.*` (per-kind happy paths, screenshot disk-write-before-enqueue, retry attempts+errorClear, cancel best-effort, internal-token gate on `workerComplete`, transactional success/failure writes)
- [ ] Manual smoke against a real Redis: deferred until PRD-126 ships a worker (no consumer to exercise the queue contract yet)

## What does NOT land

- pops-worker-food container — PRD-126
- Per-kind extraction handlers — PRDs 127–132
- AI usage logging — PRD-133
- Review queue UI consuming `list` — Epic 03

## Pre-existing test flake

`src/pages/data/__tests__/FoodDataLayout.test.tsx` fails 8 tests when the full app-food suite runs but passes 11/11 in isolation. Reproduces on `origin/main` without this PR. Not caused by this work; flagging so CI maintainers can triage separately.

### DIFF
--- a/.github/workflows/_pkg-check.yml
+++ b/.github/workflows/_pkg-check.yml
@@ -45,6 +45,7 @@ jobs:
           pnpm --filter @pops/db-types build
           pnpm --filter @pops/types build
           pnpm --filter @pops/module-registry build
+          pnpm --filter @pops/food-contracts build
           pnpm --filter @pops/app-food-db build
       - run: cd ${{ inputs.package-path }} && pnpm typecheck
         if: inputs.skip != 'true'
@@ -74,6 +75,7 @@ jobs:
           pnpm --filter @pops/db-types build
           pnpm --filter @pops/types build
           pnpm --filter @pops/module-registry build
+          pnpm --filter @pops/food-contracts build
           pnpm --filter @pops/app-food-db build
       - name: Run tests (with coverage if configured)
         if: inputs.skip != 'true'
--- a/.github/workflows/api-quality.yml
+++ b/.github/workflows/api-quality.yml
@@ -66,6 +66,7 @@ jobs:
           cd packages/db-types && pnpm build
           cd ../types && pnpm build
           cd ../module-registry && pnpm build
+          cd ../food-contracts && pnpm build
           cd ../app-food-db && pnpm build
 
       - name: Type check
--- a/.github/workflows/api-test.yml
+++ b/.github/workflows/api-test.yml
@@ -61,6 +61,7 @@ jobs:
           cd packages/db-types && pnpm build
           cd ../types && pnpm build
           cd ../module-registry && pnpm build
+          cd ../food-contracts && pnpm build
           cd ../app-food-db && pnpm build
 
       - name: Typecheck
--- a/.github/workflows/fe-quality.yml
+++ b/.github/workflows/fe-quality.yml
@@ -72,6 +72,7 @@ jobs:
           cd packages/db-types && pnpm build
           cd ../types && pnpm build
           cd ../module-registry && pnpm build
+          cd ../food-contracts && pnpm build
           cd ../app-food-db && pnpm build
 
       - name: Type check
--- a/.github/workflows/fe-test-e2e.yml
+++ b/.github/workflows/fe-test-e2e.yml
@@ -61,6 +61,7 @@ jobs:
           cd packages/db-types && pnpm build
           cd ../types && pnpm build
           cd ../module-registry && pnpm build
+          cd ../food-contracts && pnpm build
           cd ../app-food-db && pnpm build
 
       - name: Cache Playwright browsers
--- a/apps/pops-api/Dockerfile
+++ b/apps/pops-api/Dockerfile
@@ -10,6 +10,7 @@ COPY packages/types/package.json ./packages/types/
 COPY packages/module-registry/package.json ./packages/module-registry/
 COPY packages/app-food-db/package.json ./packages/app-food-db/
 COPY packages/app-lists/package.json ./packages/app-lists/
+COPY packages/food-contracts/package.json ./packages/food-contracts/
 COPY apps/pops-api/package.json ./apps/pops-api/
 
 # Install pnpm and dependencies (cached across builds)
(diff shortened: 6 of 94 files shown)


knoxio:
Addressed all 11 Copilot findings in 32d6dee1:

1. **`partialReason` persistence** — nested inside `extracted_json` on `ok: true` so `extractPartialReason` recovers it after BullMQ TTL. New test locked it in.
2. **Pre-decode base64 cap** — checks the cleaned-base64-string length against `Math.ceil((SCREENSHOT_MAX_BYTES * 4) / 3)` before `Buffer.from`.
3. **`listIngestSources` N+1** — new `fetchJobTimingsBatch` does ONE `queue.getJobs` call per request and buckets by sourceId.
4. **`food/index.ts` header** — rewritten to reflect the seven sub-routers we now mount.
5. **`deriveTitle` regex** — `/@recipe\s*\([^)]*\btitle\s*=\s*"([^"]+)"/`. Test exercises a real `@recipe(slug=carbonara, title="Spaghetti carbonara")` body.
6. **Screenshot retry mime** — `findOnDiskScreenshot` reads the per-source dir and recovers the original mime from `screenshot.{jpg,jpeg,png,webp}`. Throws when the file was evicted.
7. **`GET /screenshot` DB guard** — single prepared `SELECT archived_at FROM ingest_sources WHERE id = ?`; 404 on missing or non-null `archived_at`.
8. **`GET /video`** — same guard.
9. **`IngestStatusOutput.partialReason`** — `PartialReasonSchema` enum (mirrors `@pops/food-contracts`).
10. **`WorkerCompleteInput.ok=true.partialReason`** — same enum. New test asserts the gate rejects `'made-up-reason'`.
11. **Drizzle snapshot 0066** — reset to PRD-123's baseline; ran `drizzle-kit generate` to produce a clean 0067 snapshot containing only my `ALTER TABLE` additions. Renamed `0067_fat_deadpool` → `0067_prd_125_ingest_error_columns`.

13/13 ingest-router tests pass, full typecheck + lint + format clean.

knoxio:
R2 fixes (9 findings) landed in a2a60003:

1. **`@pops/food-contracts` runtime** — dist-built (Node16, `main: ./dist/index.js`). Build step added to `pops-shell`, `pops-api`, `pops-mcp` Dockerfiles.
2. **tsconfig** — flipped from `noEmit:true`/`bundler` to declarations + Node16 emit.
3. **`findFileWithExtension`** — preserves on-disk casing (returns `join(dir, entry)` not `join(dir, candidate)`).
4–5. **workerComplete row-existence guard** — `loadSourceRow` throws `WorkerCompleteSourceNotFound` so a missing/double-evicted `sourceId` is loud, not a silent no-op UPDATE. Both `applySuccess` and `applyFailure`.
6. **workerComplete idempotency** — `applySuccess` early-returns the existing `draft_recipe_id` when set. BullMQ retries no longer try to re-insert the slug. New test asserts single `recipes` row after two `ok:true` calls.
7. **`startIngest` rollback** — `ingest_sources` row + screenshot dir removed when enqueue throws (Redis down). Underlying error surfaces to the caller. New test asserts the count is unchanged after a failed start.
8. **`cancelIngest` state alignment** — `CANCELLABLE_STATES` is now the single source of truth used by both `findJobBySourceId`'s `getJobs` scan AND the post-fetch state check. Also fixed the spelling: BullMQ uses `waiting-children` (hyphen), not `waiting_children` (underscore) — the earlier set silently never matched. Same spelling fix applied to `ingest-state.ts`'s `BULLMQ_TO_INGEST` map.
9. **`fetchJobTimingsBatch` paging** — paginates up to `SCAN_MAX_JOBS=1000` (mirrors `removeOnComplete: { count: 1_000 }`) in 100-job pages. Early-exits when every requested sourceId is resolved. Extracted `recordJobTimings` helper to keep complexity ≤ 12.

3 new tests: workerComplete throws on missing sourceId; idempotency returns same draftRecipeId; startIngest rolls back on enqueue failure. 16/16 ingest-router tests pass; typecheck 27/27, lint clean.

PR #2853, opened 2026-06-10
test(food): PRD-146 — cook-overrides server integration tests (#2787)

knoxio:
## Summary

Closes #2787.

- Adds the dedicated `cook-overrides.test.ts` suite called out in the drift-check ticket: one assertion per row of the PRD-146 contract table
  - `kind='batch-override'` → `batch_consumptions` row written + batch decremented + FIFO batch untouched
  - `kind='external'` → no `batch_consumptions` row + audit line appended to `recipe_runs.notes`
  - `kind='partial'` → one `batch_consumptions` row + audit line appended
  - optional-line override → silently dropped (no row, no note, no decrement)
  - depleted-batch shortfall → `ShortfallUnresolved` + transaction rolled back
- Tightens `mark-cooked-overrides.ts` to honour the spec's *"skip optional lines silently (PRD-108 contract)"* clause — overrides targeting an optional line no longer write stray `batch_consumptions` rows.
- Splits the migration-boot + recipe/batch fixtures out of `cook-router.test.ts` into `cook-test-helpers.ts` (+ `cook-test-db.ts`) so both test files use the same shape. Removes the partial override coverage from `cook-router.test.ts` now that it lives in the dedicated file.

## Test plan

- [x] `pnpm exec vitest run src/modules/food/__tests__/cook-overrides.test.ts src/modules/food/__tests__/cook-router.test.ts` — 23 passed
- [x] `pnpm test` (full pops-api suite) — 4752 passed
- [x] `pnpm typecheck` (turbo, all 28 packages)
- [x] `pnpm lint` — clean
- [x] `pnpm format:check` — clean

### DIFF
--- a/apps/pops-api/src/modules/food/__tests__/cook-overrides.test.ts
+++ b/apps/pops-api/src/modules/food/__tests__/cook-overrides.test.ts
@@ -4,13 +4,11 @@
  *
  * Mirrors the spec in [PRD-146 §"Integration with PRD-144's cook
  * mutation"](docs/themes/07-food/prds/146-fifo-consumption-ui/README.md):
- * one test per row in the contract table — `batch-override`, `external`,
- * `partial`, optional-line silent skip, and the depleted-batch shortfall
- * roll-back.
- *
- * Reject-shape assertions (zero-qty bypass, wrong-variant guard) live in
- * `cook-router.test.ts` alongside the broader `MarkCookedError` matrix —
- * keeping this file focused on the happy-path persistence contract.
+ * the happy-path persistence contract (one test per row of the override
+ * matrix: `batch-override`, `external`, `partial`, optional-line silent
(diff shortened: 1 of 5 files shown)


knoxio:
Addressed all four Copilot findings in e449793d:

- **R1/R2 (fixture FK correctness)** — `seedCompiledRecipe` and `seedRecipeLine` now thread real `ingredients.id` values instead of falling back to the variant id. `SeededRecipe` now exposes `yieldIngredientId` alongside `yieldVariantId`.
- **R3 (stale header)** — header in `cook-overrides.test.ts` rewritten so it no longer points at coverage that was removed; the zero-qty and wrong-variant reject tests now live in this file.
- **R4 (qty-coverage bypass)** — `applyConsumptionOverrides` now refuses to mark a line covered unless the override's total qty (in the line's canonical unit) matches the scaled need. New tests assert the qty-mismatch and unit-mismatch reject paths.

PR #3295, opened 2026-06-15
feat(inventory): PRD-251 US-01/US-02 inventory cross-pillar denorm (audit H7)

knoxio:
## Summary

- Implements **PRD-251 US-01 + US-02** (inventory-side cross-pillar denormalisation, closes audit H7's inventory half).
- `home_inventory` gains `purchase_transaction_uri` + `owner_uri` (plus per-reference `*_stale_at` markers) as soft references shaped `pops://<pillar>/<type>/<id>`. Migration `0008_cross_pillar_uri_denorm` adds the columns + indices and promotes `purchase_transaction_id` → URI in-place.
- A nightly reconciliation worker (`apps/pops-inventory-api/src/cron/reconcile-cross-pillar.ts`) walks distinct URIs, asks the owning pillar via the typed `pillar()` proxy, and stamps `*_stale_at = now` on 404. `unavailable` retries on the next tick; bad URIs are logged for ops; rows are never deleted (existence is best-effort per PRD-251 §"Business Rules").
- Scaffolds `core.users.get` (read-only) so the cron has a stable surface to probe single-user identity (Cloudflare Access email-keyed).

## Out of scope

US-03 (finance → core denorm) lands in a separate PR per the parallelisation note in PRD-251 — no `packages/finance-db/` files touched here.

## Test plan

- [x] `pnpm --filter @pops/inventory-db typecheck/test/build` — clean (130 tests, 7 files)
- [x] `pnpm --filter @pops/pops-inventory-api typecheck/test/build` — clean (63 tests, 7 files, including the 4 PRD-251 cases: happy-path, 404, owning-pillar-unavailable, bad-URI, plus the timer scheduling)
- [x] `pnpm --filter @pops/core-db typecheck/test/build` — clean
- [x] `pnpm --filter @pops/core-api typecheck/build` — clean
- [x] `pnpm --filter @pops/api test` — 4899/4899 pass (test-utils + cross-source + per-pillar-migrations DDLs updated for new columns)
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — clean (only pre-existing warnings)
- [x] `pnpm format:check` — clean
- [x] Husky pre-commit hook passed without `--no-verify`

PR #3296, opened 2026-06-15
feat(finance-api): PRD-251 US-03 finance-side cross-pillar denorm (audit H7)

knoxio:
## Summary

Implements PRD-251 US-03 — the finance half of audit H7's cross-pillar denormalisation.

- **Schema** — `budgets` (`finance-db`) and `entities` (`core-db`) gain a nullable `owner_uri` + `owner_uri_stale_at` column pair plus a `(owner_uri)` index. Migrations: `packages/finance-db/migrations/0055_budgets_owner_uri.sql` (also mirrors the ALTER onto finance-db's local entities copy used as a FK target) and `packages/core-db/migrations/0067_entities_owner_uri.sql` (canonical for the core-owned entities table per PRD-245 US-07).
- **Reconciliation cron** — `apps/pops-finance-api/src/cron/reconcile-cross-pillar.ts` walks distinct budgets owner URIs nightly, calls `core.users.get` via the pillar SDK proxy, and marks rows stale on 404. Errors are logged and the row is preserved; bad URIs are recorded for ops. Owner-side writes only.
- **`core.users.get` scaffold** — minimal read-only procedure on `pops-core-api` that resolves `pops://core/entities/<id>` URIs to a `{ uri, displayName, kind }` projection or `null`. Same surface will serve the parallel inventory PR (US-01/02).
- **Backfill** — no legacy join column exists on `budgets` today, so the migration is column-add only; existing rows stay NULL until written. UI tolerates NULL.
- **Tests** — `apps/pops-finance-api/src/cron/__tests__/reconcile-cross-pillar.test.ts` covers happy-path / 404 marks staleAt / core unavailable / bad URI / thrown lookup / clear-on-resolve / recursive-timer scheduling against an in-memory finance.db.

## Deviation from the PRD literal

PRD-251 says "add `ownerUri` to `packages/finance-db/src/schema/entities.ts`" — that file no longer exists. PRD-245 US-07 (landed after PRD-251 was scoped) relocated the canonical `entities` schema to `@pops/core-db`. The column lives where the table lives (core-db); finance-db keeps an aligned ALTER on its FK-target copy. The PR description here flags this so it doesn't surprise a reviewer cross-referencing the PRD.

## Coordination

A separate PR (`feat/prd-251-us-01-02-inventory-denorm`) covers US-01/02 (inventory-side). Both PRs scaffold `core.users.get` — the second to land should rebase. The scaffold here is minimal (one file + one registry line) to keep the merge conflict trivial.

## Test plan

- [x] `pnpm --filter @pops/finance-db test` — 208/208 passing
- [x] `pnpm --filter @pops/core-db test` — 144/144 passing
- [x] `pnpm --filter @pops/finance-api test` — 39/39 passing (includes new cron suite)
- [x] `pnpm --filter @pops/core-api test` — all green
- [x] `pnpm typecheck` clean across the monorepo
- [x] `pnpm lint` clean (warnings only — same set as `main`)
- [x] `pnpm format:check` clean


knoxio:
**⚠️ Contract mismatch with #3295** — flagged during rebase coordination.

After rebase, the two PRs share a `core.users.get` scaffold but the **shapes don't agree**:
- #3295 (inventory) router: input `{ email }`, output `{ data: { email } }`
- #3296 (finance) cron caller: input `{ uri }`, expects `{ uri, displayName, kind } | null`

Tests pass because both stub the lookup. At runtime, finance cron will hit a 400 against inventory's router. The type system can't catch it because `CoreRouter` is `AnyTRPCRouter` (opaque).

**Recommended fix as a follow-up PR (do NOT block these two)**:
1. Standardise on the URI shape (per PRD-244 typed-proxy convention): `get({ uri }) → { uri, displayName, kind, ownerEmail? } | null`
2. Update the inventory cron caller to use the new shape
3. Re-run finance cron in integration test mode

Both PRs are safe to land independently because the cron only runs nightly and is best-effort — a 400 logs and retries next tick. But the follow-up should be a tracked TODO.

cc the IMPLEMENTATION-PLAN refresh.

PR #3317, opened 2026-06-15
chore(core-api): drop POPS_INTERNAL_API_KEY from pillar registry endpoints

knoxio:
## Summary

Drops the shared `POPS_INTERNAL_API_KEY` requirement from the pillar registry endpoints (`/core.registry.{register,heartbeat,deregister}`). Per ADR-027, the docker network is the trust boundary — the shared key added no real security since anything reachable on those endpoints is already inside the compose network.

## Changes

- `register.ts` / `heartbeat.ts` / `deregister.ts` — drop `resolveApiKey`, shared-key gates, and per-row `apiKeyHash` checks
- `register-helpers.ts` / `heartbeat-helpers.ts` — drop `apiKey` from body parsers
- `auth.ts` — deleted (`sha256Hex` + `constantTimeEquals` no longer used)
- `app.ts` / `handlers.ts` — drop `resolveApiKey` wiring through `CoreApiDeps`
- `register.ts` — drop `rejectReservedPillarId` so in-tree pillar IDs can self-register (PRD-250)
- `packages/core-db/src/services/pillar-registry.ts` — `apiKeyHash` now `string | null`; external registrations write `null`
- `packages/core-db/src/schema/pillar-registry.ts` — docstring updated; column stays nullable for backward compat (no migration)
- Tests updated to drop the auth/env-key/key-rotation cases; new happy-path covers in-tree pillar registration returning 200

## Out of scope

- The legacy `pops-api` server-SDK auth surface (separate concern; still uses `POPS_INTERNAL_API_KEY`)
- Dropping the `api_key_hash` column — left nullable to avoid migration churn
- `homelab-infra` changes — will be a no-op once this lands

## Test plan

- [x] `pnpm --filter @pops/core-api test` — 130 passing
- [x] `pnpm --filter @pops/pillar-sdk test` — 553 passing
- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm lint:boundaries`
- [x] `pnpm --filter @pops/api typecheck`


knoxio:
After merge, capivara will need core-api + 7 pillar-apis recreated via docker compose up -d. The POPS_API_KEY_FILE secret mount on pops-api/pops-worker can stay (still used for that surface) but stops being functional for registry registration.

PR #3319, opened 2026-06-15
feat(pops-api): PRD-248 US-05c flip watch-history mixed-tx writes to cerebrum SDK

knoxio:
## Summary

Option D burn-down for the `media/watch-history` slice of PRD-248 US-05. The watch-history fan-out into cerebrum debrief now goes through the `pillar('cerebrum').debrief.*` SDK; the in-monolith `@pops/cerebrum-db` import and the `cerebrum/debrief/router` import both leave the handlers/ subtree.

- `handlers/cerebrum-fan-out.ts` — `pillar('cerebrum').debrief.logWatchCompletion`, fire-and-forget on the write path. Helper stays sync-returning so `logWatch` keeps its sync signature; flipping it to `async` would cascade through the out-of-scope `plex/` / `arr/` / `rotation/` callers. `PillarCallError` + generic failures are logged via the structured `logger.warn` and swallowed — writer is idempotent on `(watchHistoryId, mediaType, mediaId)`; self-heal on next completion or via the deferred reconciler.
- `handlers/query-helpers.ts` — `deleteWatchHistoryEntry` is now `async`; it awaits `pillar('cerebrum').debrief.deleteByWatchHistoryId(...)` BEFORE the media-side delete commits. Cerebrum failure aborts the delete so the user can retry. `watch-history/router.ts` `delete` procedure awaits the new async signature.
- Tests: `service.test.ts` and `media/debrief/debrief.test.ts` rewired to mock `@pops/pillar-sdk/server`; cover happy path, partial failure, idempotent retry (write path), and abort-on-SDK-failure (delete path).
- `.dependency-cruiser-known-violations.json` shrinks: 3 entries removed (`cerebrum-fan-out → cerebrum/debrief/router`, `service.test → cerebrum/debrief/router`, `query-helpers → @pops/cerebrum-db`).
- PRD-248 US-05 doc updated: `log-watch-event` + `query-helpers` ticks flipped; README marks US-05 partial. Remaining sites under `media/comparisons/lib/*` and `media/debrief/{service,queue-status}.ts` belong to sibling slices.

## Scope guard

Strictly limited to `apps/pops-api/src/modules/media/watch-history/handlers/`. The fire-and-forget pattern is deliberate: it avoids the async cascade that would otherwise reach `plex/`, `arr/`, and `rotation/` (already-flipped sub-trees per the PRD-248 US-05c scope contract).

## Test plan

- [x] `pnpm --filter @pops/api typecheck` clean
- [x] `pnpm --filter @pops/api test` (4936/4936 pass)
- [x] `pnpm --filter @pops/api build` clean
- [x] `pnpm typecheck` clean
- [x] `pnpm lint` clean
- [x] `pnpm lint:boundaries` clean
- [x] Husky pre-commit + pre-push passed without `--no-verify`


knoxio:
Heads up: a debrief rip-out is in flight (per user decision to nuke the feature rather than widen the SDK — see PRD-248 stall pattern). The watch-history fan-out + delete-side-effect that this PR flips will be **removed** by the rip-out. This PR is correct as-is and proves the SDK path works for the mixed-tx case, but its diff will be subsumed.

Merge order:
- Land rip-out first → close this PR (work is moot but doc-proven)
- Or land this first → rip-out rebases + the removed-debrief diff also drops these calls

knoxio:
Superseded by #3321 (debrief feature rip-out). The watch-history fan-out + delete-side-effect this PR flipped have been removed entirely.

PR #3518, opened 2026-06-22
refactor(federation): fold storybook into libs/ui dev surface [P2-T04]

knoxio:
## P2-T04 — Fold `@pops/storybook` into `libs/ui` as a dev surface

Storybook is `@pops/ui`'s dev tool, not a separate deployable package. After this, `@pops/storybook` no longer exists.

### What moved
- `apps/pops-storybook/.storybook/` → `libs/ui/.storybook/` (story globs + Vite aliases re-pathed from the stale `packages/` layout to `libs/ui/src` and `pillars/*/app/src`).
- `apps/pops-storybook/scripts/check-app-package-coverage.mjs` → `libs/ui/scripts/check-storybook-coverage.mjs`, retargeted to assert every pillar frontend has a Vite alias (the drift signal that now matters), wired into `@pops/ui` `test`/`test:coverage`.
- Storybook tooling (`storybook`, `@storybook/addon-a11y`, `@chromatic-com/storybook`) + the CodeMirror/Lezer deps the DslEditor story needs → `libs/ui` **devDependencies**. Added `storybook` + `build-storybook` scripts.
- `git rm -r apps/pops-storybook` (standalone package gone).
- Deleted `.github/workflows/storybook-quality.yml` — `ui-quality.yml` already builds `@pops/ui` and runs its test/coverage, so the dev surface is covered there. Leaving it would hard-fail on `main` (its `_pkg-check` would `cd` into a dead path).

### Reject-the-naive-fold (DEPENDENCY-DAG §D6)
`libs/ui` takes **no** `@pops/app-*` in `dependencies`, `peerDependencies`, **`devDependencies`**, `exports`, or `main`. The dev surface consumes the frontends through **Vite source aliases** in `.storybook/main.ts`, which resolve `@pops/app-*` directly to each pillar's `app/src` — so no workspace edge is needed.

A `ui → app-*` package edge was rejected for two reasons: every `@pops/app-*` already depends on `@pops/ui`, so the edge forms a **turbo `^build` cycle** (breaks `pnpm typecheck`), and it would put a pillar in the lib's graph (federation isolation break). Because the aliases already handle resolution, the edge buys nothing.

Consequently **no `ALLOWED_DEV_PILLAR_DEPS` entry was needed** — the lib→pillar guard passes with zero pillar deps on `libs/ui`. The guard's allowlist doc records *why* `@pops/ui` is intentionally absent, and a guard test pins that a naively re-added `app-*` devDep would still be flagged.

### Verify (all green locally)
- `pnpm -r list` — no `@pops/storybook` package.
- `libs/ui` runtime deps / `exports` / `main` carry no `@pops/app-*` (nor do devDeps).
- `node scripts/ci/check-lib-no-pillar-import.mjs` → exit 0; self-test passes; 9 guard unit tests pass.
- `node libs/ui/scripts/check-storybook-coverage.mjs` → aliases all 7 frontends.
- `pnpm --filter @pops/ui typecheck` + `test`/`test:coverage` pass.
- `pnpm lint`, `pnpm format:check`, `npx yaml-lint .github/workflows/*.yml`, whole-repo `pnpm typecheck` (43/43, no cycle) all pass.
- No stale `apps/pops-storybook` / `@pops/storybook` references in yml/json/ts.


knoxio:
@coderabbitai full review

PR #3541, opened 2026-06-23
feat(shell): registry-driven boot install-set [P7-T03]

knoxio:
## What

Re-points the shell's boot install-set from the build-time `MODULES`/`INSTALLED_MODULES` constants to the **live registry snapshot**, making the registry the source of truth for which pillars mount (federation north-star, RD-3). This is a boot-architecture change: the shell previously built its router + app-rail synchronously at module-eval time; this introduces the async boot boundary that did not exist before.

## How

- **Async boot boundary** — `main.tsx` now `await`s `fetchBootRegistry()` before `createRoot().render()`. While the fetch is in flight it renders a minimal unbranded splash; the router is built **after** the snapshot resolves.
- **`buildRouter(manifests)`** — `router.tsx` no longer exports a module-eval `router` constant; `App.tsx` builds it once (memoised) from the boot-resolved manifests.
- **`BootRegistryProvider`** — a sibling to `PillarStatusProvider`. The nav consumers (`AppRail`, `Sidebar`, `PageNav`, `RootLayout`, `IndexRedirect`) read `registeredApps` from `useRegisteredApps()` instead of the module-eval constant.
- **`boot-snapshot.ts`** — `resolveBootRegistry(snapshot)` maps the snapshot onto the install set; `walkRegistry` is reused **byte-for-byte unchanged**.
- **Single registry client** — the snapshot fetch + parse is shared between the Settings UI and boot via `registry-snapshot-fetch.ts` (no third client).

## Resilience contract (never brick)

- snapshot non-empty → the registry's `registered` entries **are** the install set (in-repo via the static bundle map, external pillars via the runtime loader, backend-only dropped — `walkRegistry`'s existing decision tree).
- snapshot empty / fetch failed / timed out → fall back to the **static bundle-map floor**, narrowed by the runtime install shim (`isInstalledModule`) so the build-time `POPS_APPS` contract still governs the offline path. The shell always renders its in-repo app set.

## Tests

- `registry-walk.test.ts` green and untouched (the pinned invariant; `walkRegistry` is byte-identical).
- `installed-modules.test.ts` rewritten off the stale `'reads from MODULES'` assertion; covers `bootEntries` / `staticFloorEntries` / `filterAppManifests`.
- `boot-snapshot.test.ts` (new) — resilience: fetch-ok (in-repo + external), fetch-fail, non-OK, empty, and timeout all assert the never-brick fallback.
- `registry-snapshot-fetch.test.ts` (new) — canonical URL + parse + soft-fail.
- shell typecheck + 605 unit tests + gen:nginx:check + build + dep-cruiser boundaries all green.

## Notes / drift from the plan

- The plan's RD-3 text said the snapshot type comes from `@pops/types`; the correct type is `PillarSnapshot` from `@pops/pillar-sdk` (the `@pops/types` shape is the minimal `{id,baseUrl}` projection). The full snapshot is fetched at `/registry-api/registry/pillars`.
- The fallback floor honours `isInstalledModule` (not the raw bundle map): removing it would break the pinned finance-only install-set e2e, whose narrowing rides the build-time `POPS_APPS` contract and has no live registry in the e2e environment. `MODULES` (the superset) is gone from the install-set path; the runtime install shim governs only the offline floor.
- RD-9 (`KnownPillarId`→`string`) is intentionally NOT done — it is gated after this.

---

## Fix round (adversarial review + orchestrator)

**M1 — never-brick zero-UI hole (closed).** `resolveBootRegistry` gated on `snapshot.length > 0`. A *non-empty* snapshot whose pillars are all backend-only (no bundle-map hit, no `assetsBaseUrl`) — the live state mid-deploy on a host restart, before app pillars re-register — resolved to `manifests=[]`/`registeredApps=[]` with `source='registry'`, never consulting the floor: an app-less shell, reachable on a real capivara restart. Fixed: the resolver now falls back to the static floor when a live snapshot resolves to **zero mountable UI** (not just zero array length), degrading exactly as if the registry were unreachable. New test feeds an all-backend-only (`registry`+`orchestrator`) snapshot and asserts `source==='static-floor'` + the full in-repo rail.

**M2 — online-path coverage.**
- **(a)** `boot-online-render.test.tsx` drives the real online pipeline (stubbed `fetch` → `fetchBootRegistry` → `BootRegistryProvider` → a rail consumer) on a non-empty snapshot with an in-repo + an external pillar, asserting `source==='registry'` and a non-blank rendered rail (both pillars, external nav synthesized). It runs in the **unit/fe-quality CI gate that actually executes on PRs**.
- **(b) — honest gap.** The Playwright e2e suite (`fe-test-e2e.yml`) is `workflow_dispatch`-only: it was written against the deleted `apps/pops-api` tRPC monolith and cannot pass without a full REST rewrite, so **nothing in it runs in PR CI**. Adding a `/registry-api` route-stub spec there would be dead coverage — it would never guard against the prod-only regression M2 describes. The real guard is therefore the rendered jsdom test in (a). What *was* wired for the online path: a `/registry-api` dev proxy in `vite.config.ts` so the dev boot fetch reaches the registry pillar instead of 404ing into the floor. **Follow-up:** when the e2e suite is rewritten against the REST stack (`TODO(e2e-rest-rewrite)`), add a Playwright `route`/MSW stub spec serving a non-empty snapshot fixture and assert the registry-driven rail mounts.

**Folds.** `main.tsx`: `bootstrap().catch(...)` now mounts the static floor if resolution ever throws, and `<App>` is wrapped in an `ErrorBoundary` — never-brick is structural, not reliant on `fetchBootRegistry` being total. `railBundleMap`: `synthesizeExternalBundleEntry` wrapped in the same `try/catch` the router-side `resolveExternalManifest` uses, so a bad external descriptor can't throw out of boot resolution on the rail path (symmetric with the walk path).

**IndexRedirect.** The optimistic landing target now derives from the first **live** app (`registeredApps[0]`) instead of a `/finance` literal, so a finance-less registry doesn't flash `NotInstalledPage`. The `data`-present `/settings` path (operator excluded every live app via `POPS_APPS`) is unchanged. New tests cover a finance-less rail landing on `/media`.

**Known limitation (acknowledged, not changing now).** The router is built once per session (immutable, `useMemo` on the boot-resolved manifests). For a LAN boot-once shell that is acceptable; a mid-session router rebuild on registry change is out of scope.

**Local CI to green before push.** Build graph warmed (`mise run build`), then: shell typecheck clean, full shell suite **612 tests pass**, oxfmt `--check` clean, oxlint exit 0 (pre-existing warnings only), `gen:nginx:check` up to date, shell `build` ok, dep-cruiser boundaries — no new violations (82 known, unchanged), root `tsc -b --noEmit` (pre-push gate) clean. `walkRegistry` byte-identical; `registry-walk.test.ts` untouched + green.

### DIFF
--- a/pillars/shell/src/app/App.tsx
+++ b/pillars/shell/src/app/App.tsx
@@ -5,15 +5,18 @@ import { isNetworkError } from '@/lib/network-error';
 import { useThemeStore } from '@/store/themeStore';
 import { MutationCache, QueryCache, QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 import { RouterProvider } from 'react-router';
 import { toast } from 'sonner';
 
 import { PillarSdkProvider } from '@pops/pillar-sdk/react';
 import { Toaster, TooltipProvider } from '@pops/ui';
 
+import { BootRegistryProvider } from './BootRegistryProvider';
 import { PillarStatusProvider } from './pillars';
-import { router } from './router';
+import { buildRouter } from './router';
+
+import type { BootRegistry } from './boot-snapshot';
 
 const NETWORK_ERROR_TOAST_ID = 'network-down';
 
@@ -42,8 +45,21 @@ const queryClient = new QueryClient({
   mutationCache: new MutationCache({ onError: notifyNetworkError }),
 });
 
-export function App() {
+interface AppProps {
+  /**
+   * The boot-resolved install set (P7-T03). Resolved in `main.tsx` from the
+   * live registry snapshot (or the static floor when the registry is
+   * unreachable) before first render, then threaded in here.
+   */
+  readonly bootRegistry: BootRegistry;
+}
+
+export function App({ bootRegistry }: AppProps) {
   const theme = useThemeStore((state) => state.theme);
+
+  // The install set is fixed for the lifetime of a session: the snapshot is
+  // resolved once at boot, so the router is built once from those manifests.
+  const router = useMemo(() => buildRouter(bootRegistry.manifests), [bootRegistry.manifests]);
 
   // Apply theme class to root element
   useEffect(() => {
@@ -53,11 +69,13 @@ export function App() {
   return (
     <QueryClientProvider client={queryClient}>
       <PillarSdkProvider>
-        <PillarStatusProvider>
-          <TooltipProvider>
-            <RouterProvider router={router} />
-          </TooltipProvider>
-        </PillarStatusProvider>
+        <BootRegistryProvider value={bootRegistry}>
+          <PillarStatusProvider>
+            <TooltipProvider>
+              <RouterProvider router={router} />
+            </TooltipProvider>
+          </PillarStatusProvider>
+        </BootRegistryProvider>
         {!import.meta.env.VITE_E2E && <ReactQueryDevtools initialIsOpen={false} />}
         <Toaster />
       </PillarSdkProvider>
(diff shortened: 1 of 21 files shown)


knoxio:
@coderabbitai review

